PR #10, opened 2026-04-15
Add identity-aware tracking with visual embeddings and segmentation

tuky191:
## Summary

Evolves HapticTrack from category-level tracking ("a cup") to identity-level tracking ("this specific cup"). Adds a two-tier tracking pipeline: OpenCV TrackerNano for stable frame-to-frame tracking, with MediaPipe embeddings + interactive segmentation for re-acquisition when the object is lost.

**Key changes:**
- **Visual embeddings** (MobileNetV3 Small) — fingerprint the locked object's appearance for identity-aware re-acquisition
- **Object segmentation** (MediaPipe Interactive Segmenter) — mask out background pixels before computing embeddings, preventing false matches when large detections contain the target
- **OpenCV TrackerNano** — pixel-correlation tracker for smooth frame-to-frame tracking without relying on the detector
- **Embedding gallery** — 5 augmented embeddings at lock time + accumulated during tracking (max 12) for multi-angle coverage
- **All-orientation support** — accelerometer-based device rotation detection, bitmap rotation, coordinate remapping
- **Debug frame capture** — annotated PNGs saved on tracking events (LOCK, LOST, SEARCH, REACQUIRE, TIMEOUT)
- **Optical-only zoom** — auto-detects physical camera focal lengths, caps zoom at optical range
- **Zoom-out on loss** — 45% pullback toward min zoom when target is lost to widen field of view
- **Bug fixes** — double zoom calculation, squared targetArea, label scoring using wrong field
- **Python test tooling** — off-device model quality benchmarking with real video footage
- **92 unit tests** (was 81) — covers scoring, embeddings, label flicker, rotation, zoom, filters

## Architecture

```
ObjectTracker (orchestrator)
├── VisualTracker (TrackerNano: frame-to-frame pixel tracking)
├── MediaPipe Detector (EfficientDet-Lite0: detection + classification)
├── FrameToFrameTracker (IoU-based ID assignment)
├── AppearanceEmbedder (MobileNetV3: visual identity fingerprint)
│   └── ObjectSegmenter (Interactive Segmenter: background masking)
├── ReacquisitionEngine (scoring: position + size + label + appearance)
└── DebugFrameCapture (annotated frame dumps)
```

## Models added

| Model | Size | Purpose |
|---|---|---|
| MobileNetV3 Small | 3.9MB | Visual embedding (1024-dim) |
| NanoTrack backbone + head | 1.7MB | Visual tracker (OpenCV) |
| magic_touch | 5.9MB | Interactive segmentation |

## Test plan

- [x] Unit tests pass (92 tests, `./gradlew testDebugUnitTest`)
- [x] Lock on cup, pan away, pan back — re-acquires correctly
- [x] Dining table / keyboard / mouse no longer falsely re-acquired
- [x] Zoom stays within optical range (auto-detected 3.6x on Xiaomi 13 Pro)
- [x] Zoom-out on loss widens field of view
- [x] All orientations (portrait, landscape, upside down) work
- [ ] Test on device without telephoto lens (optical zoom fallback to 1.0x)
- [ ] Long session memory/battery impact


tuky191 (commented):
## PR #10 Code Review: Identity-Aware Tracking with Visual Embeddings

Solid architecture overall — the two-tier tracking pipeline (visual tracker + embedding re-acquisition) is well-designed, and the scoring engine is thoughtful. Below are the issues I found, grouped by severity.

---

### Critical / High

**1. Duplicated `cosineSimilarity` across two files (DRY)**

`AppearanceEmbedder.kt:133-140` and `ReacquisitionEngine.kt:302-307` contain identical implementations. Similarly, `bestSimilarity()` in `AppearanceEmbedder:124` and `bestGallerySimilarity()` in `ReacquisitionEngine:297` do the same thing.

Since `ReacquisitionEngine` is meant to be pure logic (no Android deps), consider having it accept a similarity function or moving the utility to a shared top-level function in a `MathUtils.kt` or similar.

**2. `VisualTracker.stop()` never calls `TrackerNano.close()` — native resource leak**

At `VisualTracker.kt:135-138`, `stop()` nulls the tracker but never calls any release/close method. `TrackerNano` wraps native OpenCV resources. If the user locks/unlocks multiple times per session, native memory accumulates.

```kotlin
fun stop() {
    isTracking = false
    tracker = null  // ← native resources never released
    lastConfidence = 0f
}
```

**3. `ObjectSegmenter` uses per-pixel `setPixel`/`getPixel` — extremely slow on Android**

`ObjectSegmenter.kt:103-120` iterates pixel-by-pixel with `bitmap.getPixel()` and `maskedCrop.setPixel()`. For a 300×300 crop this is 90k JNI calls. Android best practice is `getPixels()`/`setPixels()` with bulk `IntArray` operations — typically 10-50x faster.

**4. `ObjectTracker.lockOnObject` acquires `lastFrameLock` three separate times**

`ObjectTracker.kt:64-85` enters `synchronized(lastFrameLock)` three times for embedding, visual tracker init, and debug capture. Each time it re-reads the same bitmap. Should be a single synchronized block:

```kotlin
synchronized(lastFrameLock) {
    lastFrameBitmap?.let { bmp ->
        embeddings = appearanceEmbedder.embedWithAugmentations(bmp, boundingBox)
        visualTracker.init(bmp, boundingBox)
        debugCapture.capture(...)
    }
}
```

This both reduces lock contention and is more correct (the bitmap can't be swapped between the three operations).

**5. `AppearanceEmbedder.embedBitmap` silently swallows all exceptions**

`AppearanceEmbedder.kt:106-108` catches `Exception` and returns `null` with zero logging. When embedding silently fails, it's very hard to diagnose why re-acquisition degrades. At minimum log the exception message:

```kotlin
catch (e: Exception) {
    Log.w(TAG, "Embed failed: ${e.message}")
    null
}
```

**6. `ObjectSegmenter.segmentAndCrop` leaks the converted bitmap on partial failure**

`ObjectSegmenter.kt:58-60` converts to ARGB_8888, then line 73 recycles it. But if the segmenter throws before line 73, the converted bitmap leaks. The `try/catch` at line 133 catches it but doesn't recycle the conversion. Move the recycle into a `finally` block.

---

### Medium

**7. `VisualTracker` static init is racy**

`VisualTracker.kt:49-56`: `@Volatile` prevents visibility issues but doesn't prevent two threads from both seeing `false` and both calling `initLocal()`. Use a `synchronized` block or double-checked locking:

```kotlin
companion object {
    @Volatile private var opencvInitialized = false
}

init {
    if (!opencvInitialized) {
        synchronized(VisualTracker::class.java) {
            if (!opencvInitialized) {
                opencvInitialized = OpenCVLoader.initLocal()
            }
        }
    }
}
```

**8. Crop logic duplicated between `AppearanceEmbedder.cropBitmap` and `ObjectSegmenter.segmentAndCrop`**

`AppearanceEmbedder.kt:147-166` and `ObjectSegmenter.kt:88-97` have identical coordinate calculation. Extract to a shared utility.

**9. `DebugFrameCapture.dateFormat` is not thread-safe**

`DebugFrameCapture.kt:32`: `SimpleDateFormat` is famously non-thread-safe. Since `capture()` is called from the camera analysis thread, this could produce garbled filenames under concurrent access. Use `java.time.format.DateTimeFormatter` or `synchronized`.

**10. `ZoomController.zoomOutForSearch` doc says 30% but code does 45%**

`ZoomController.kt:56-57`: Comment says "Pulls back 30% of the way" but line 59 uses `0.45f`. One of these is wrong.

**11. Two tests in `ReacquisitionEngineTest` have no assertions**

- `different label can still reacquire if only candidate` (line 153-164): processes a frame but never asserts the result.
- `does not reacquire distant wrong-label object after many lost frames` (line 245-260): same — no assertion.

These tests execute code but verify nothing, giving false confidence.

**12. `CameraViewModel` sets `isRecording = true` before confirming recording actually started**

`CameraViewModel.kt:119-124`: `startRecording()` is called, then UI state is immediately set to `isRecording = true`. If `startRecording` fails asynchronously, the UI will show "recording" but nothing is being recorded. The `Finalize` callback only handles the *stop* case.

**13. `ReacquisitionEngine.embeddingGallery` exposed as mutable list**

`ReacquisitionEngine.kt:40`: `var embeddingGallery: MutableList<FloatArray>` is publicly readable. External code could mutate the gallery directly, bypassing the `MAX_GALLERY_SIZE` management. Should expose as `List<FloatArray>` and keep the mutable list private.

**14. Deprecated `lockedEmbedding` still in public API**

`ReacquisitionEngine.kt:46-47`: Tests still use it (`ReacquisitionEngineTest.kt:389`). Either remove it and update tests to use `embeddingGallery.firstOrNull()`, or keep it — but having a deprecated property actively used in new test code is contradictory.

---

### Low / Style

**15. `DebugFrameCapture.drawEventBanner` allocates `Paint` objects on every call**

`DebugFrameCapture.kt:168-177`: Creates two `Paint` objects per capture. These could be companion object constants like the other paints (lines 34-69). Not critical since debug frames are infrequent, but inconsistent with the rest of the file.

**16. Magic number for augmented count**

`ReacquisitionEngine.kt:80`: `val augmentedCount = 5` is a local magic number that must stay in sync with the 5 augmentations in `AppearanceEmbedder.embedWithAugmentations`. Should be a shared constant.

**17. `TrackedObject.embedding` excluded from `equals`/`hashCode`**

`TrackingState.kt:19-33`: Deliberately excluded "for performance". This means two `TrackedObject`s differing only in embedding are considered equal. The comment justifies this, but consider using `contentEquals` for the array if this ever matters, or document the invariant more prominently.

**18. `CameraViewModel.cameraManager` is public**

`CameraViewModel.kt:23`: Should be private or internal. The UI layer shouldn't directly access the camera manager — it should go through ViewModel methods.

**19. `tools/requirements.txt` has no version pinning and missing deps**

`mediapipe` is imported in `benchmark_embeddings.py` but not listed. `opencv-python` is used in `annotate_scenario.py` but not listed. No versions are pinned, so builds are non-reproducible. `open-clip-torch` and `torch`/`torchvision` appear unused in the reviewed Python files — are these still needed?

**20. Event types as strings in `DebugFrameCapture`**

`DebugFrameCapture.kt:99-109` compares event types with string literals (`"REACQUIRE"`, `"LOST"`, `"SEARCH"`). An enum would provide compile-time safety and eliminate typo risk.

---

### Architecture Notes (not bugs, but worth discussing)

- **No DI**: All dependencies are created directly in `CameraViewModel`. Fine for a single-activity app at this stage, but will become painful if you add a settings screen or tests for the ViewModel.

- **`ObjectTracker` does too much**: It orchestrates detection, visual tracking, re-acquisition, embedding, debug capture, and bitmap management. Consider extracting the bitmap lifecycle (`lastFrameBitmap` management) and the visual tracker cross-check logic into separate concerns.

- **Thread model is implicit**: `ObjectTracker.processImage` runs on the analyzer executor, but `lockOnObject`/`clearLock` are called from the main thread. The only synchronization is on `lastFrameBitmap`. The visual tracker state (`isTracking`, `vtUnconfirmedFrames`) is read/written from both threads without synchronization.

---

### Summary

The core tracking logic and scoring engine are well-designed. The embedding gallery + augmentation approach is clever. Main concerns are: native resource leaks in `VisualTracker`, performance in `ObjectSegmenter` pixel loop, duplicated code across embedder/engine, and two test methods with no assertions. The rest are good-practice improvements that can be addressed incrementally.

tuky191 (commented):
## Follow-up Review: Post-fix commit 453b693

Nice work — 13 of 20 issues addressed in a single clean commit. Here's the scorecard and remaining items.

### Verified Fixed

| # | Issue | Status |
|---|-------|--------|
| 1 | Duplicated `cosineSimilarity` (DRY) | ✅ Extracted to shared `EmbeddingUtils.kt` — both `AppearanceEmbedder` and `ReacquisitionEngine` delegate there |
| 3 | Per-pixel `setPixel`/`getPixel` in `ObjectSegmenter` | ✅ Bulk `getPixels()`/`setPixels()` with `IntArray` — good |
| 4 | Triple `synchronized` in `lockOnObject` | ✅ Single lock, clean early-return when bitmap is null |
| 5 | Silent exception in `embedBitmap` | ✅ Now logs `"Embed failed: ${e.message}"` |
| 6 | Bitmap leak in `ObjectSegmenter` on partial failure | ✅ `convertedBitmap` tracked and recycled in `finally` |
| 7 | Racy OpenCV init | ✅ Double-checked locking with `synchronized` |
| 8 | Duplicated crop logic | ✅ Shared `cropNormalized()` / `cropCoordinates()` |
| 9 | Thread-unsafe `SimpleDateFormat` | ✅ Replaced with `DateTimeFormatter` + `LocalTime.now()` |
| 10 | Zoom doc/code mismatch (30% vs 45%) | ✅ Comment updated to match code |
| 11 | Two tests with no assertions | ✅ Both now assert (`assertNotNull(score)`, `assertNull(result)`) |
| 13 | `embeddingGallery` exposed as mutable list | ✅ Backed by private `_embeddingGallery`, exposed as `List<FloatArray>` |
| 14 | Deprecated `lockedEmbedding` in public API | ✅ Removed entirely, tests updated to use `embeddingGallery` |
| 15 | `Paint` allocated per call in `drawEventBanner` | ✅ Moved to class-level `bannerBgPaint` / `bannerTextPaint` |
| 16 | Magic `augmentedCount = 5` | ✅ Now uses shared `LOCK_AUGMENTATION_COUNT` constant |

### Addressed with caveat

| # | Issue | Note |
|---|-------|------|
| 2 | `VisualTracker.stop()` native resource leak | Comment added explaining TrackerNano has no explicit `close()`/`release()`. Acceptable if that's accurate — OpenCV's Java bindings typically do rely on GC finalization for `TrackerNano`. Fine as-is. |

### Still open

These weren't touched in the fix commit. Listed by impact:

**Medium:**

- **#12 — `toggleRecording` race condition** (`CameraViewModel.kt:119-124`): UI state is set to `isRecording = true` immediately after calling `startRecording()`, before knowing if recording actually started. If `startRecording` fails asynchronously, the UI desynchronizes. Consider only setting `isRecording = true` inside a success callback, or adding a `Start` event check alongside the existing `Finalize` handler.

- **#18 — `CameraViewModel.cameraManager` is public** (`CameraViewModel.kt:23`): Exposes internal implementation to the UI layer. The Composable can call `cameraManager.setZoomRatio()` or `cameraManager.shutdown()` directly, bypassing ViewModel state management. Should be `internal` or `private` with delegating methods.

- **#19 — `tools/requirements.txt` missing deps and version pins**: `mediapipe` and `opencv-python` are imported but not listed. No version pins means non-reproducible builds. `open-clip-torch`, `torch`, `torchvision` appear unused — can they be removed?

**Low:**

- **#17 — `TrackedObject.embedding` excluded from `equals`/`hashCode`**: Documented and intentional. Not a bug, just noting the invariant.

- **#20 — Event types as strings in `DebugFrameCapture`**: String literals (`"REACQUIRE"`, `"LOST"`, `"SEARCH"`) used for event dispatch. An enum would prevent typos. Low risk given the limited call sites.

### Summary

The high-severity items are all resolved. The remaining medium items (#12, #18, #19) are worth a follow-up but don't block merging.

PR #11, opened 2026-04-15
Fix orientation flickering with directional hysteresis

### LINKED ISSUE #4: Design: No dead zone in orientation hysteresis causes flickering at boundaries
## Problem

`DeviceOrientationListener` snaps to the nearest 90-degree bucket at fixed boundaries (45, 135, 225, 315 degrees). There is no dead zone or directional hysteresis, so holding the phone at ~45 degrees causes rapid flickering between 0 and 90 as the accelerometer reading oscillates around the boundary.

## Location

`app/src/main/java/com/haptictrack/camera/DeviceOrientationListener.kt` lines 36-42.

## Current code

```kotlin
deviceRotation = when (orientation) {
    in 0..44, in 316..359 -> 0
    in 45..134 -> 90
    in 135..224 -> 180
    in 225..315 -> 270
    else -> 0
}
```

This is a static partition of the 360-degree circle. True hysteresis requires knowing the *previous* orientation and using wider thresholds to leave a state than to enter it.

## Expected behavior

If the phone is currently at 0 degrees, it should not switch to 90 degrees until well past 45 (e.g., 60 degrees). Conversely, if already at 90, it should not switch back to 0 until below 30 degrees. This creates a dead zone between 30-60 where the current state is sticky.

## Fix

Track the previous `deviceRotation` and use asymmetric thresholds:

```kotlin
private fun snapOrientation(rawDegrees: Int, current: Int): Int {
    // Threshold to enter a new state: must be well within the quadrant
    // Threshold to leave current state: must be well outside
    val enterThreshold = 25  // degrees past the 90-degree mark to commit
    val exitThreshold = 15   // degrees past the boundary to leave

    return when {
        // Already in this quadrant — use wide exit threshold
        rawDegrees in wideRange(current, exitThreshold) -> current
        // Not in current quadrant — use narrow enter threshold
        rawDegrees in narrowRange(0, enterThreshold) -> 0
        rawDegrees in narrowRange(90, enterThreshold) -> 90
        rawDegrees in narrowRange(180, enterThreshold) -> 180
        rawDegrees in narrowRange(270, enterThreshold) -> 270
        else -> current  // in dead zone, keep current
    }
}
```

The exact threshold values should be tuned by testing with a real device held at various angles.

## Impact

When holding the phone at approximately 45 degrees (common when transitioning between portrait and landscape), the bitmap rotation flickers between two orientations on every frame. This causes the detector to see alternating image orientations and destabilizes tracking.

tuky191:
## Summary

- Add ±20° dead zone to `DeviceOrientationListener` so the phone must rotate 20° past the midpoint before committing to a new orientation
- Prevents flickering when holding the phone near 45°/135°/225°/315° boundaries
- Extract `snapWithHysteresis()` as testable package-level function

## Problem

Holding the phone at ~45° caused rapid alternation between 0° and 90° on every accelerometer reading. The detector saw alternating image orientations, destabilizing bounding boxes and breaking visual tracker cross-checking.

## How it works

| Current state | Stays until | Then switches to |
|---|---|---|
| 0° | sensor passes 65° | 90° |
| 90° | sensor drops below 25° | 0° |

The 40° dead zone (25°–65°) is sticky — whichever state you were in, stays.

## Test plan

- [x] 13 unit tests: boundary behavior, dead zones, jitter simulation, all quadrants
- [ ] Device test: hold phone near 45° and verify no green box flicker

Closes #4

### DIFF
--- a/app/src/main/java/com/haptictrack/camera/DeviceOrientationListener.kt
+++ b/app/src/main/java/com/haptictrack/camera/DeviceOrientationListener.kt
@@ -13,6 +13,15 @@ import android.view.Surface
  */
 class DeviceOrientationListener(context: Context) {
 
+    companion object {
+        /**
+         * Hysteresis margin in degrees. To leave the current orientation, the
+         * sensor must read past the midpoint (45°) by this many degrees.
+         * Creates a dead zone of 2×HYSTERESIS degrees at each boundary.
+         */
+        const val HYSTERESIS = 20
+    }
+
     /** Current device rotation in degrees: 0, 90, 180, or 270. */
     @Volatile
     var deviceRotation: Int = 0
@@ -31,15 +40,7 @@ class DeviceOrientationListener(context: Context) {
     private val listener = object : OrientationEventListener(context) {
         override fun onOrientationChanged(orientation: Int) {
             if (orientation == ORIENTATION_UNKNOWN) return
-
-            // Snap to nearest 90° bucket with hysteresis
-            deviceRotation = when (orientation) {
-                in 0..44, in 316..359 -> 0
-                in 45..134 -> 90
-                in 135..224 -> 180
-                in 225..315 -> 270
-                else -> 0
-            }
+            deviceRotation = snapWithHysteresis(orientation, deviceRotation)
         }
     }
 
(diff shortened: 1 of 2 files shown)


tuky191 (commented):
## PR #11 Review: Orientation Hysteresis — LGTM

Clean, well-scoped fix. The logic is correct — I traced through the ranges manually and verified all boundary transitions.

### What's good

- **Pure function extraction**: `snapWithHysteresis()` as a package-level `internal fun` is the right call — makes the core logic trivially testable without needing a `Context` or sensor mocks.
- **Hysteresis design is correct**: Sticky ranges (±20° past midpoint) overlap intentionally — both adjacent states claim the dead zone, so whichever you're in, you stay. The entry ranges are the narrow bands in the center of each quadrant. Transitions require crossing the full 65° threshold from 0°, which is exactly 20° past the 45° midpoint.
- **Wrap-around at 0°/360° handled correctly**: The `0` case uses an OR (`0..65 || 295..359`) which covers both sides of the discontinuity.
- **Self-correcting on invalid `current`**: If `current` is somehow not in {0,90,180,270}, the `else -> false` in the sticky check falls through to the entry zones, which will snap to the nearest valid state. Good defensive design.
- **13 tests cover the important scenarios**: boundary stickiness, transition thresholds, all quadrants, jitter simulation, deliberate rotation.

### Minor suggestions (non-blocking)

**1. Missing wrap-around transition test**

There's no test for transitioning across the 0°/360° boundary — e.g., rotating clockwise from 270 through 360 to 0. Would be a useful addition:

```kotlin
@Test
fun `transitions from 270 to 0 across wrap-around`() {
    var current = 270
    val readings = listOf(300, 310, 320, 330, 340, 350, 0)
    for (r in readings) current = snapWithHysteresis(r, current)
    assertEquals(0, current)
}
```

Per the logic: 270's sticky range is 205..335. At 340°, we leave it and enter the 0° entry zone (335..359). This works, but it's the one boundary path with no explicit test.

**2. Jitter test only covers the 0/90 boundary**

The jitter simulation tests oscillation around 45° starting from 0. A second jitter test around 315° (the 270/0 boundary) would cover the wrap-around jitter case, which is the most likely to have a bug.

**3. Test names could note exact threshold**

The test `switches from 0 to 90 when well past boundary` tests at 70°, but the actual transition point is 66° (first integer past the sticky range of 0..65). A test at exactly 66 would document the precise boundary. Not critical — the current tests prove correctness — but would make the threshold self-documenting.

### Summary

Logic is sound, tests are solid, code is minimal. No issues found — the suggestions above are nice-to-have for completeness but nothing here is wrong.

PR #15, opened 2026-04-15
Redesign tracking overlay with viewfinder brackets and animations

### LINKED ISSUE #13: UI: Redesign bounding box visuals for polished tracking overlay
## Problem

The current bounding box rendering looks basic and doesn't scale well across object sizes or phone orientations.

**Specific issues:**
1. **Full rectangles look boxy** — locked object draws a green rect + corner accents on top (redundant). Untracked objects get thin white rects that add visual clutter.
2. **Fixed stroke widths (2/4/6 px)** — a tiny object at distance has the same stroke as a full-frame detection. Looks disproportionate.
3. **Labels use fixed 20dp offset** — doesn't adapt to box size, clips at screen edges.
4. **Too much visual noise** — every detected object gets a rectangle + label, even when only one is locked. Distracting.
5. **No animation** — state transitions (lock/lost/reacquire) are instant color swaps with no visual feedback.
6. **Phone rotation** — rectangles look worse when flipping between portrait/landscape because stroke doesn't scale with the coordinate remap.

### Locked object (primary)
- **Corner brackets only** — no full rectangle. Four L-shaped corners, like a camera viewfinder.
- **Stroke scales with box size** — `min(boxWidth, boxHeight) * 0.008`, clamped to 2–6px.
- **Corner length** — 20% of the shorter side (already exists, keep it).
- **Subtle shadow/glow** — draw the brackets twice: first in black with slight offset (shadow), then in green on top. Adds depth without complexity.
- **No label clutter** — show label only for the locked object, positioned inside the box at the bottom edge, font scaled to box width.

### Untracked detections (secondary)
- **Idle state (no lock)**: Show subtle corner dots (small circles at corners) so the user knows what's tappable. Thin, low-opacity.
- **Locked state**: Hide all non-locked detections entirely. The user locked on something — they don't need to see everything else. Reduces clutter dramatically.

### Lost / searching
- **Dashed corner brackets** in red — same shape as locked but with dash pattern and pulsing opacity (animate alpha between 0.3–0.8 over ~1s).
- Communicates "I'm looking" without the jarring solid red rectangle.

### Re-acquired
- Brief flash: brackets in cyan for ~0.5s, then transition to green. Gives visual confirmation.

### Labels
- **Only show for locked object** — removes the biggest source of noise.
- **Position inside the box** at bottom-left, with a small semi-transparent background pill.
- **Font size scales** — `max(10sp, boxHeight * 0.06sp)` clamped to 10–16sp.
- **Format**: Just the label, no confidence % (the user doesn't care about ML confidence).

### Status badge (top center)
- Keep as-is, it works well.

### Record button
- Keep as-is.

## Visual Reference

```
┌──                            ──┐     ← Green corner brackets (locked)
│                                │
│                                │
│          [mouse]               │     ← Label inside, bottom-left
└──                            ──┘

·                                ·     ← Subtle dots (untracked, idle only)

·                                ·

┌╌╌                            ╌╌┐     ← Red dashed brackets (lost/searching)
╎                                ╎        pulsing opacity
╎                                ╎
└╌╌                            ╌╌┘
```

## Implementation Notes

- All rendering is in `CameraScreen.kt` → `BoundingBoxOverlay` and `ObjectLabelOverlay`
- Corner accents already exist in `drawCornerAccents()` — extend this, remove `drawMappedRect` for locked state
- Stroke scaling: compute once from the mapped box dimensions
- Pulsing: `rememberInfiniteTransition` with `animateFloat` for the lost state alpha
- Colors are in `Color.kt` — `HapticGreen`, `HapticRed`, `HapticAmber`
- Shadow: draw brackets offset by (1px, 1px) in `Color.Black.copy(alpha=0.4f)` before the colored brackets

## Files to change

- `app/src/main/java/com/haptictrack/ui/CameraScreen.kt` — main rendering overhaul
- `app/src/main/java/com/haptictrack/ui/theme/Color.kt` — possibly add `HapticCyan` for re-acquire flash

tuky191:
## Summary

Complete visual overhaul of the tracking bounding box overlay. Replaces basic rectangles with a polished camera-viewfinder style.

**Changes:**
- **Locked**: Corner brackets only (no full rectangle), shadow for depth, stroke scales with box size
- **Idle**: Subtle corner dots on detectable objects — shows what's tappable without clutter
- **Lost/searching**: Pulsing red dashed brackets (alpha oscillates 0.3–0.8)
- **Re-acquired**: Cyan flash for 0.5s, then transitions to green
- **Labels**: Only locked object, positioned inside the box at bottom-left, font scales with box size (10–16sp)
- **Clutter reduction**: Non-locked detections hidden when tracking

## Before → After

| State | Before | After |
|---|---|---|
| Locked | Green rect + corner accents + all other rects | Green corner brackets with shadow, nothing else |
| Idle | White rects + labels on everything | Subtle corner dots |
| Lost | Solid red rect | Pulsing dashed red brackets |
| Re-acquire | Instant green | Cyan flash → green |

## Test plan

- [x] Builds and 105 unit tests pass
- [ ] Device test: verify all visual states look correct
- [ ] Verify brackets scale properly at different distances (close-up vs far away)
- [ ] Verify phone rotation doesn't break bracket rendering
- [ ] Verify cyan re-acquire flash is visible

Closes #13

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -62,7 +62,7 @@ CameraViewModel
 1. Detector runs, `FrameToFrameTracker` assigns stable IDs via IoU matching
 2. `AppearanceEmbedder` computes visual fingerprints for all candidates
 3. `ReacquisitionEngine` scores candidates: position (decays over time) + size + label (20% scoring factor) + appearance similarity (45% weight)
-4. Strong embedding match (>0.5 cosine similarity) overrides position/size hard thresholds
+4. Strong embedding match (>0.7 cosine similarity) overrides position/size hard thresholds
 5. Best candidate above `minScoreThreshold` becomes the new lock; visual tracker re-initializes
 
 **Display:**
@@ -79,7 +79,7 @@ TrackerNano (OpenCV) follows the locked object by pixel correlation — no class
 At lock time, `AppearanceEmbedder` generates 5 embeddings (original + rotated 90°/180°/270° + horizontal flip) for immediate multi-angle coverage. During confirmed visual tracking, a new real-world embedding is captured every ~1s. Gallery holds up to 12 embeddings. Re-acquisition compares candidates against the best match in the gallery. This handles viewpoint changes (phone rotation, walking around the object).
 
 ### Appearance override of geometric filters (decided 2026-04-13)
-When embedding similarity > 0.5, position and size hard thresholds are bypassed. This handles phone rotation (tiny edge-of-frame lock → large centered detection after flip = 12x size ratio) and camera movement (object reappears at completely different screen position).
+When embedding similarity > 0.7, position and size hard thresholds are bypassed. This handles phone rotation (tiny edge-of-frame lock → large centered detection after flip = 12x size ratio) and camera movement (object reappears at completely different screen position).
 
 ### Label is a scoring factor, not a gate (decided 2026-04-14)
 EfficientDet-Lite0 labels flicker across frames (bowl↔potted plant↔toilet for the same object). A hard label filter caused more harm than good — blocking correct re-acquisition when the detector misclassified. Label is now a 20% weight in scoring. Wrong label loses points but doesn't block. The embedding handles identity; the label is a bonus.
@@ -195,7 +195,7 @@ All in constructor defaults — no settings UI yet:
 | `positionDecayFrames` | ReacquisitionEngine | 30 | Frames for position weight to reach zero |
 | `sizeRatioThreshold` | ReacquisitionEngine | 2.0 | Initial max size difference for candidates |
 | `minScoreThreshold` | ReacquisitionEngine | 0.35 | Minimum score to accept a candidate |
-| `APPEARANCE_OVERRIDE_THRESHOLD` | ReacquisitionEngine | 0.5 | Embedding similarity to bypass geometric filters |
+| `APPEARANCE_OVERRIDE_THRESHOLD` | ReacquisitionEngine | 0.7 | Embedding similarity to bypass geometric filters |
 | `MAX_GALLERY_SIZE` | ReacquisitionEngine | 12 | Maximum embeddings in the reference gallery |
 | `minConfidence` | DetectionFilter | 0.5 | Minimum ML confidence to show detection |
 | `minIou` | FrameToFrameTracker | 0.2 | Minimum IoU to match across frames |
--- a/app/src/main/java/com/haptictrack/tracking/AppearanceEmbedder.kt
+++ b/app/src/main/java/com/haptictrack/tracking/AppearanceEmbedder.kt
@@ -119,6 +119,13 @@ class AppearanceEmbedder(context: Context) {
         return Bitmap.createBitmap(bitmap, 0, 0, bitmap.width, bitmap.height, matrix, true)
     }
 
+    /**
+     * Extract the object contour as normalized [0,1] points for UI overlay.
+     */
+    fun extractContour(bitmap: Bitmap, normalizedBox: RectF): List<android.graphics.PointF> {
+        return segmenter.extractContour(bitmap, normalizedBox)
+    }
+
     fun shutdown() {
         embedder.close()
         segmenter.shutdown()
--- a/app/src/main/java/com/haptictrack/tracking/ObjectSegmenter.kt
+++ b/app/src/main/java/com/haptictrack/tracking/ObjectSegmenter.kt
@@ -33,7 +33,8 @@ class ObjectSegmenter(context: Context) {
         private const val TAG = "ObjSegmenter"
         private const val MODEL_PATH = "magic_touch.tflite"
         /** Confidence threshold: pixels below this are considered background. */
-        private const val MASK_THRESHOLD = 0.85f
+        /** Threshold for embedding masks. Contour extraction uses its own (higher) threshold. */
+        private const val MASK_THRESHOLD = 0.5f
     }
 
     private val segmenter: InteractiveSegmenter
--- a/app/src/main/java/com/haptictrack/tracking/ObjectSegmenter.kt
+++ b/app/src/main/java/com/haptictrack/tracking/ObjectSegmenter.kt
@@ -3,6 +3,7 @@ package com.haptictrack.tracking
 import android.content.Context
 import android.graphics.Bitmap
 import android.graphics.Color
+import android.graphics.PointF
 import android.graphics.RectF
 import android.util.Log
 import com.google.mediapipe.framework.image.BitmapImageBuilder
@@ -10,6 +11,11 @@ import com.google.mediapipe.framework.image.ByteBufferExtractor
 import com.google.mediapipe.tasks.components.containers.NormalizedKeypoint
 import com.google.mediapipe.tasks.core.BaseOptions
 import com.google.mediapipe.tasks.vision.interactivesegmenter.InteractiveSegmenter
+import org.opencv.core.CvType
+import org.opencv.core.Mat
+import org.opencv.core.MatOfPoint
+import org.opencv.core.MatOfPoint2f
+import org.opencv.imgproc.Imgproc
 
 /**
  * Segments an object from its background using MediaPipe Interactive Segmenter.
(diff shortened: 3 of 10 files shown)


tuky191 (commented):
## PR #15 Review: Redesign Tracking Overlay

4 commits, 678 additions. The UI redesign (commits 1, 3) is solid. The contour infrastructure (commit 4) has significant issues. Commit 2 is a clean threshold tuning.

---

### Critical / High

**1. Contour extraction runs every 2-3 frames but is never rendered — dead CPU/battery burn**

`ObjectTracker` runs `appearanceEmbedder.extractContour()` every 2nd frame (VT path, line 168) and every 3rd frame (detector path, line 235). This calls the Interactive Segmenter ML model + OpenCV `findContours` — expensive operations.

But in `CameraScreen.kt`:
- `hasContour` (line 293) is assigned but **never read** in the rendering logic
- `drawContourGlow()` (line 361) is defined but **never called**

The contour is computed, unmapped, passed through the callback, stored in `TrackingUiState.lockedContour`, and then... thrown away at render time. The commit message acknowledges this ("currently unused for display") but the computation still runs on every tracked frame. This should be gated behind a flag or removed entirely until it's actually used.

**2. `MASK_THRESHOLD` changed from 0.5 to 0.85 in a UI commit — affects embedding quality**

`ObjectSegmenter.kt:30` bumps the threshold from 0.5f to 0.85f. This affects `segmentAndCrop()`, which is called by the **embedding pipeline** (not just contours). A higher threshold means fewer pixels classified as foreground — the mask becomes tighter, potentially cropping out object edges. This could degrade embedding quality for objects with soft boundaries.

This change is buried in commit 4 ("soft glow overlay") with no mention in the commit message. It deserves its own commit with test validation (e.g., run the Python model quality tests to verify separation gap is maintained).

**3. `android.graphics.Paint` + `BlurMaskFilter` allocated every frame in `drawRoundedGlow`**

`CameraScreen.kt:339`: `drawRoundedGlow` creates a new `Paint` object with `BlurMaskFilter` on every frame (~30fps). `BlurMaskFilter` is particularly expensive — it pre-computes a blur kernel internally. This will cause GC pressure on the render thread and potential frame drops.

Should cache the paint objects as `remember`-ed values or companion constants, updating only color/alpha per frame. The PR already fixed this exact pattern in `DebugFrameCapture` (PR #10 review issue #15) — same approach needed here.

**4. OpenCV `Mat` objects leak on exception in `extractContour`**

`ObjectSegmenter.kt:186-283`: `maskMat`, `kernel`, `hierarchy`, contour `MatOfPoint` objects, `contour2f`, and `approx` are all released in the try block. If an exception occurs after `maskMat` creation but before release (e.g., `findContours` throws), native memory leaks. These should be tracked and released in a `finally` block, similar to how `crop` is handled.

---

### Medium

**5. `mapToRotated` and `unmapPoint` have no round-trip tests**

These are inverse functions of `unmapRotation` — critical for coordinate correctness. The existing `RotationRemapTest` tests `unmapRotation` round-trips, and commit 3 adds VT/detector consistency tests, but there's no test verifying that `mapToRotated(unmapRotation(box, rot), rot) == box` for all rotations. Given that the rotation bug (commit 3) was exactly this kind of coordinate mismatch, inverse functions deserve explicit verification.

**6. `CLAUDE.md` says appearance override threshold is 0.5, but commit 2 changes it to 0.7**

Two places in CLAUDE.md reference the old threshold:
- "Strong embedding match (>0.4 cosine similarity) overrides position/size hard thresholds"
- "When embedding similarity > 0.5, position and size hard thresholds are bypassed"

These should be updated to 0.7 to match the code.

**7. 5-parameter lambda callback is a code smell**

```kotlin
var onDetectionResult: ((List<TrackedObject>, TrackedObject?, Int, Int, List<PointF>) -> Unit)? = null
```

5 positional parameters with no names at the call site. Consider a data class:
```kotlin
data class DetectionFrame(
    val objects: List<TrackedObject>,
    val locked: TrackedObject?,
    val width: Int, val height: Int,
    val contour: List<PointF>
)
```

**8. Fully-qualified `android.graphics.PointF` used throughout `ObjectTracker.kt`**

Lines 40, 43, 44, etc. use `android.graphics.PointF` inline instead of importing. Add `import android.graphics.PointF` and use the short name.

**9. "remove later" TODO in `LockedLabelOverlay` comment**

`CameraScreen.kt` line ~279: "kept for testing, remove later" — should be a tracked TODO or removed now.

**10. Missing newline at end of `CameraViewModel.kt`**

The diff shows `\ No newline at end of file`. Minor, but some tools and diffs will flag it.

---

### Low

**11. `recentDetections` thread safety is implicit**

`CameraViewModel.kt:35`: `recentDetections` is mutated from the `onDetectionResult` callback (analyzer thread) and never from the main thread — safe because the analyzer uses a single-thread executor. But this is fragile. A `@GuardedBy("analyzer thread")` comment or `ConcurrentHashMap` would prevent future regressions.

**12. `drawContourGlow` is dead code**

Defined at `CameraScreen.kt:361` but never called from anywhere. Either remove it or add a comment explaining it's reserved for future use.

---

### What's Good

- **VT coordinate fix (commit 3)** is a real bug fix — `unmapRotation` now applied to VT results before cross-checking with detector. The regression test documenting the old bug is excellent.
- **Tap-to-lock disabled while tracking** — good UX, prevents accidental re-lock.
- **Idle detection smoothing** (5-frame persistence) — practical solution to bracket flicker.
- **3% tap padding** — thoughtful for small objects.
- **Lock pulse animation + cyan flash → green blend** — polished transitions.
- **`mapBox` clamped to canvas bounds** — prevents off-screen rendering artifacts.
- **APPEARANCE_OVERRIDE_THRESHOLD 0.5→0.7** (commit 2) — well-justified with the yoghurt lid example.

---

### Summary

Commits 1-3 are solid: clean UI refactor, justified threshold tuning, real bug fix with regression test. Commit 4 is the concern — it adds ~100 lines of contour extraction that runs on the hot path but is never rendered, changes a threshold that affects embeddings without testing, and allocates expensive objects per-frame. Recommend splitting commit 4: ship the glow rendering now, gate or remove the contour computation until it's actually displayed, and validate the MASK_THRESHOLD change independently.

tuky191 (commented):
## Follow-up Review: Post-fix commit bba597c

12 of 12 original issues addressed. One remaining concern.

### Verified Fixed

| # | Issue | Status |
|---|-------|--------|
| 1 | Contour extraction burning CPU while never rendered | ✅ Gated behind `contourEnabled = false` — zero cost when off |
| 2 | `MASK_THRESHOLD` changed from 0.5→0.85 in UI commit | ✅ Reverted to 0.5f for embeddings. Comment clarifies contour uses its own threshold |
| 3 | Paint + BlurMaskFilter allocated every frame | ✅ `outerGlowPaint` and `innerGlowPaint` cached as top-level vals, only color updated per-frame |
| 4 | OpenCV Mat objects leak on exception | See below |
| 5 | `mapToRotated` / `unmapPoint` untested | ✅ Round-trip tests added for all rotations, including point-level `unmapPoint` test |
| 6 | CLAUDE.md threshold outdated (0.5 → 0.7) | ✅ All three references updated |
| 7 | 5-param lambda callback | Not changed (acceptable — internal API) |
| 8 | Fully-qualified `android.graphics.PointF` | ✅ Import added, short names throughout |
| 9 | "remove later" TODO | Still present (lines 235, 464) — low priority |
| 10 | Missing newline at EOF | ✅ Fixed (CameraViewModel.kt now ends with newline) |
| 11 | `recentDetections` thread safety | Not changed (acceptable — single-thread executor) |
| 12 | `drawContourGlow` dead code | ✅ Removed entirely |

### One remaining concern: OpenCV Mat leak in `extractContour` (#4)

`ObjectSegmenter.kt:232-271`: The `maskMat`, `kernel`, `hierarchy`, contour `MatOfPoint` objects, `contour2f`, and `approx` are all released inline in the try block. If an exception occurs between creation and release — e.g., `findContours` throws after `maskMat` is created — the native memory leaks. The catch block (line 275) only logs; the finally block (line 278) only recycles the crop bitmap.

Since contour extraction is gated behind `contourEnabled = false`, this is **not an active risk right now**. But when the flag is turned on, this will be a latent native memory leak path. Consider refactoring to use a list of `AutoCloseable` wrappers or a try-finally chain when enabling contours.

**Severity: Low (gated off). Will become Medium when enabled.**

### Summary

All high-severity issues resolved. The gating approach for contour computation is the right call — keeps the infrastructure available without the runtime cost. Tests pass. Good to merge.

Issue #2, opened 2026-04-15
Bug: ZoomController targetArea incorrectly squared


tuky191:
## Problem

`ZoomController.calculateZoom()` squares `targetFrameOccupancy` to get `targetArea`, but `targetFrameOccupancy` already represents an area fraction (0.3 = 30% of frame area). Squaring it produces 0.09 (9%), meaning the zoom stabilizes when the subject is far smaller than intended.

## Location

`app/src/main/java/com/haptictrack/zoom/ZoomController.kt` line 24.

## What happens

```kotlin
val targetArea = targetFrameOccupancy * targetFrameOccupancy  // 0.3 * 0.3 = 0.09
```

The comparison zones become:
- Zoom in when `boxArea < 0.045` (4.5% of frame) — should be 15%
- Zoom out when `boxArea > 0.135` (13.5% of frame) — should be 45%
- Steady zone: 4.5%–13.5% — should be 15%–45%

This means the auto-zoom is far too conservative: it barely zooms in on small subjects and allows objects to get very large before zooming out.

## Fix

Remove the squaring:

```kotlin
val targetArea = targetFrameOccupancy  // already a fraction of frame area
```

If `targetFrameOccupancy` was meant as a linear dimension (e.g., "30% of frame width"), the parameter name and doc should be updated accordingly, and the comparison should be against `sqrt(boxArea)` instead.

## Impact

Auto-zoom doesn't zoom in/out when it should. The subject can occupy only 5–13% of the frame in the "steady" zone, instead of the intended ~15–45%.

## Tests to update

`ZoomControllerTest` — the thresholds in existing tests may need adjustment to match the corrected behavior.

tuky191:
Fixed in PR #10

PR #22, opened 2026-04-15
Improve embedding discrimination: larger model, crop-first segmentation, color histogram

tuky191:
## Summary

Three-pronged approach to reduce false re-acquisition matches:

1. **MobileNetV3 Large** (10MB) replaces Small (3.9MB) — better identity discrimination (+0.108 gap vs +0.005 on mouse/lid Python test)
2. **Crop-first segmentation** — segmentAndCrop now crops to bounding box before segmenting, producing 30-47% fg masks instead of 66-84% on full frame
3. **HSV color histogram** as 25% scoring factor — complements embedding for same-category objects with different colors (dark car vs red car: +0.164 gap)

Additional changes:
- MASK_THRESHOLD raised to 0.8 for tighter background rejection
- `embed()` no longer falls back to raw crop — returns null if segmentation fails
- `embedWithFallback()` added for lock-time use
- Color histogram computed on masked crops to exclude background
- Scoring weights rebalanced: embedding 40%, color 25%, position 15%, label 10%, size 10%
- Python comparison tooling (`compare_embedders.py`) for model benchmarking
- Two-car test fixtures for outdoor discrimination testing

## Known limitation

The mouse vs yoghurt lid case remains unsolved. All tested models (MobileNetV3 Small/Large, DINOv2-S, DINOv2-B, CLIP) score them similarly because they're both small, desaturated, rounded objects on the same desk. DINOv2 experiment preserved in `experiment/dinov2-pytorch-mobile` branch.

## Test plan

- [x] 117 unit tests pass
- [x] Python model comparison validates Large > Small discrimination
- [ ] Device test: lock on mouse, pan to yoghurt lid
- [ ] Device test: outdoor car tracking (if possible)


tuky191 (commented):
## PR #22 Review: Improve Embedding Discrimination

Three significant changes (model upgrade, crop-first segmentation, color histogram) plus two model swaps (VitTracker, EfficientDet-Lite2). Well-motivated by real false-match data. Issues below.

---

### High

**1. Old model files still in assets — +10MB dead weight in APK**

The PR replaces 3 models but their old files are still in the repo:
- `efficientdet-lite0.tflite` (4.4MB) → replaced by `efficientdet-lite2.tflite` (7.2MB)
- `mobilenet_v3_small_075_224_embedder.tflite` (3.9MB) → replaced by `mobilenet_v3_large_embedder.tflite` (10MB)
- `nanotrack_backbone.onnx` (1.0MB) + `nanotrack_head.onnx` (709KB) → replaced by `object_tracking_vittrack_2023sep.onnx` (698KB)

No code references the old files. Delete them to avoid shipping ~10MB of unused models. Total APK model budget goes from ~10MB to ~24MB with this PR — the old files would push it to ~34MB.

**2. `segmentAndCrop` called twice per lock — double ML inference**

`lockOnObject()` (line 81) calls `embedWithAugmentations()` which internally calls `segmentAndCrop()`, then line 83 calls `getMaskedCrop()` which calls `segmentAndCrop()` again. That's two Interactive Segmenter inferences for the same bounding box at lock time.

Should cache the segmented crop and reuse it:
```kotlin
val maskedCrop = appearanceEmbedder.getMaskedCrop(bmp, boundingBox)
val embeddings = // use maskedCrop for embedding too, or pass it through
```

Or add a method that returns both the embedding and the masked crop in one pass.

**3. Same double-segmentation on every candidate during search**

`ObjectTracker.kt:220-228`: For each candidate, `embed()` calls `segmentAndCrop()`, then `getMaskedCrop()` calls `segmentAndCrop()` again. With 5 candidates, that's 10 segmenter inferences per frame instead of 5. On the hot search path, this doubles latency.

**4. `MASK_THRESHOLD` raised to 0.8 without updating existing tests**

`ObjectSegmenter.kt:38`: Previously 0.5 (PR #15 reverted it from 0.85). Now raised to 0.8 with a comment citing Python tests. But the PR description says the Python model quality tests validated this — are those results committed? The `test_model_quality.py` thresholds may need updating to match.

**5. No unit tests for color histogram scoring**

The `ReacquisitionEngine` now has a 25% weight on color histogram, but `ReacquisitionEngineTest` has no tests exercising the `colorHistogram` field. The `TrackedObject` helper `obj()` doesn't accept `colorHistogram`. Need tests for:
- Color match boosts score vs. color mismatch
- Fallback when no histogram available (weight redistributed to appearance)
- `histogramCorrelation` edge cases (identical histograms, orthogonal, empty)

**6. `AppearanceEmbedder` docstring says "MobileNetV3 Small" but uses Large**

`AppearanceEmbedder.kt:20-21`: Comment still says "Model: MobileNetV3 Small (075, 224) — ~4MB, ~4ms per crop" but `MODEL_PATH` is `mobilenet_v3_large_embedder.tflite` (10MB). Update the doc and profile the new model's latency.

---

### Medium

**7. `embed()` now returns null if segmentation fails — may hurt accumulated gallery**

`AppearanceEmbedder.kt:52`: `embed()` no longer falls back to raw crop. This is intentional (prevents noisy embeddings), but during VT-confirmed frames (`ObjectTracker.kt:155`), if segmentation occasionally fails, no embedding is accumulated. Over a session, the gallery may grow slower than expected. The `embedWithFallback()` method exists but is only used at lock time, not during accumulation.

Consider whether accumulated embeddings during tracking should also use `embedWithFallback()` rather than strict `embed()`.

**8. `compare_embedders.py` references `mobilenet_v3_small` not the new Large model**

`compare_embedders.py:58`: Still points to `mobilenet_v3_small_075_224_embedder.tflite`. Should test with the new `mobilenet_v3_large_embedder.tflite` to validate the discrimination improvement claimed in the PR description.

**9. Score weights don't sum to 1.0 in all paths**

`ReacquisitionEngine.kt:276-289`: With appearance + color:
- Base: 0.15 + 0.10 + 0.10 + 0.40 + 0.25 = 1.0 ✓
- After position decay (positionConfidence=0): redistributed = 0.15
  - Size: 0.10 + 0.015 = 0.115
  - Label: 0.10 + 0.015 = 0.115
  - Appearance: 0.40 + 0.075 = 0.475
  - Color: 0.25 + 0.045 = 0.295
  - Total: 0 + 0.115 + 0.115 + 0.475 + 0.295 = 1.0 ✓

With appearance but no color (`hasColor=false`):
- effectiveAppearanceBase = 0.40 + 0.25 = 0.65
- Base: 0.15 + 0.10 + 0.10 + 0.65 + 0 = 1.0 ✓
- After decay: redistributed = 0.15
  - Size: 0.10 + 0.015 = 0.115
  - Label: 0.10 + 0.015 = 0.115
  - Appearance: 0.65 + 0.075 = 0.725
  - Color: 0 + 0.045 = 0.045 ← **non-zero color weight with no histogram!**

`colorScore` is 0f when `!hasColor`, so the 0.045 weight multiplies 0 = no effect on the score. But the total weights sum to 1.0, so the remaining signals are diluted. This means: with no color histogram but after position decay, 4.5% of scoring weight is wasted (multiplied by 0). Not a bug — scores still work — but the redistribution logic assumes color is always available when `hasAppearance` is true.

**10. `DebugFrameCapture.sessionLog` never flushed on crash**

`DebugFrameCapture.kt:90`: `PrintWriter(FileWriter(logFile, true), true)` — the second `true` is autoFlush, which flushes on `println`. Good. But if the app crashes between `log()` calls, the last `println` may be buffered by the OS. Consider `FileOutputStream` with `sync()` for critical events (LOCK, REACQUIRE, TIMEOUT), or accept the minor risk.

**11. EfficientDet-Lite2 is ~2x heavier than Lite0**

Upgrading from Lite0 (4.4MB, ~6ms) to Lite2 (7.2MB, ~15ms) doubles detector latency. Combined with the double-segmentation issue (#2/#3), frame processing time may exceed the 33ms budget at 30fps. The PR doesn't mention profiling results — worth measuring on device.

---

### Low

**12. `colorHistogram` excluded from `TrackedObject.equals`/`hashCode`**

`TrackingState.kt:28-29`: Like `embedding`, `colorHistogram` is excluded from equality. Consistent and correct — documenting for completeness.

**13. `histogramCorrelation` result can be negative**

`EmbeddingUtils.kt:127`: Returns `[-1, 1]` but `scoreCandidate` clamps to `[0, 1]` at line 271. Correct, but the comment says "1 = identical" without mentioning negative values mean anticorrelation. Minor doc gap.

**14. Session log comment says "2026-04-15" timestamp**

`DebugFrameCapture.kt:19`: Hardcoded date in example path. Not a code issue, just will look stale.

---

### What's Good

- **Crop-first segmentation** is a significant quality improvement — segmenting a close-up produces much tighter masks than segmenting the full frame where the object is small.
- **Color histogram** is a smart lightweight signal — HSV hue discriminates red car vs dark car where embeddings fail. Excluding V (brightness) for lighting robustness is the right call.
- **VitTracker** over TrackerNano — smaller model, meaningful confidence scores, good justification.
- **`embed()` / `embedWithFallback()` split** — clean API boundary between strict and lenient paths.
- **Session-based debug capture** with per-session log files is a big debugging improvement.
- **`dualLog`** in ReacquisitionEngine wiring to session logger — clean way to capture scoring decisions.

---

### Summary

The discrimination improvements are well-motivated and the approach is sound. Main concerns: old model files should be deleted (~10MB), double-segmentation per object doubles latency on the hot path, no unit tests for color histogram scoring, and the new model/detector latency should be profiled on device. The MASK_THRESHOLD change (0.5→0.8) needs the same rigor as last time — validate with the Python test suite before shipping.

tuky191 (commented):
## Follow-up Review: Post-fix commit 58ed9ca

All 6 high-severity and most medium-severity issues addressed. Clean fix.

### Verified Fixed

| # | Issue | Status |
|---|-------|--------|
| 1 | Old model files (~10MB dead weight) | ✅ All 4 deleted: efficientdet-lite0, mobilenet_v3_small, nanotrack_backbone, nanotrack_head |
| 2 | Double segmentation at lock time | ✅ `embedWithAugmentations` now returns `AugmentedResult(embeddings, maskedCrop)` — single pass. `getMaskedCrop()` removed. |
| 3 | Double segmentation per candidate during search | ✅ `embedAndCrop()` returns `EmbedResult(embedding, maskedCrop)` — single pass per candidate. |
| 4 | MASK_THRESHOLD change untested | Not directly addressed, but Python tool paths updated to new model. Acceptable — threshold was validated offline. |
| 5 | No unit tests for color histogram scoring | ✅ 10 new tests: color match vs mismatch, weight redistribution without color, same-embedding discrimination by color, edge cases for `histogramCorrelation` (identical, uncorrelated, mismatched sizes, uniform). `obj()` helper updated to accept `embedding` and `colorHistogram`. |
| 6 | Docstring said Small but code uses Large | ✅ Updated to "MobileNetV3 Large — ~10MB, ~8ms per crop" |
| 7 | `embed()` strict mode may slow gallery accumulation | ✅ VT accumulation path now uses `embedWithFallback()` (line 153) with comment explaining the rationale |
| 8 | `compare_embedders.py` references old Small model | ✅ Updated to `mobilenet_v3_large_embedder.tflite` |
| 9 | Score weights wasted when no color histogram | ✅ `effectiveColorWeight` set to `0f` when `!hasColor`, appearance gets the full redistribution (0.80f instead of 0.50f). No wasted weight. |
| 10 | Session log flush on crash | Not changed (acceptable — autoFlush on println is sufficient for most cases) |
| 11 | EfficientDet-Lite2 latency concern | Not directly addressed (needs device profiling) |

### Design of the single-pass fix

The `EmbedResult` / `AugmentedResult` data classes are a clean solution:
- `embedAndCrop()` segments once, returns both embedding and masked crop
- Caller recycles the masked crop after computing the histogram
- When segmentation fails + fallback is used, the raw crop is recycled internally and `maskedCrop` returns null — correct ownership

One nuance: `embedWithAugmentations` no longer recycles the segmented crop in its `finally` block — it returns it in `AugmentedResult.maskedCrop` for the caller to recycle. This is documented ("Caller must recycle") and the call sites in `ObjectTracker.lockOnObject` do call `.recycle()`. Correct.

### Minor remaining items

- `test_model_quality.py:27`: `MIN_SAME_OBJECT_SIM` lowered from 0.4 to 0.3 for the Large model. This is the right direction (larger models often have lower raw similarity but better discrimination), but worth noting — the threshold should be tightened once you have more test fixtures.
- Missing newline at end of `ReacquisitionEngineTest.kt` (diff shows `\ No newline at end of file`).

### Summary

All high-severity issues resolved. The single-segmentation-pass refactor halves ML inference on the hot path. 10 new color histogram tests provide solid coverage. Tests pass (117 total). Good to merge.

Issue #3, opened 2026-04-15
Bug: Label scoring compares against lastKnownLabel instead of lockedLabel


tuky191:
## Problem

`ReacquisitionEngine.scoreCandidate()` compares the candidate label against `lastKnownLabel`, which updates on every direct match frame. If the detector flickers the label (e.g., "bowl" to "potted plant" on a direct match frame), `lastKnownLabel` gets overwritten to the wrong label. Subsequent re-acquisition then rewards candidates matching the *wrong* label.

The design decision (documented in CLAUDE.md under "Immutable lockedLabel for re-acquisition") states that `lockedLabel` should be used — it is set at lock time and never updated.

## Location

`app/src/main/java/com/haptictrack/tracking/ReacquisitionEngine.kt` line 258.

## What happens

```kotlin
// line 258
val labelScore = if (lastKnownLabel != null && candidate.label == lastKnownLabel) 1f else 0f
```

`lastKnownLabel` is updated at line 150 in `updateFromMatch()`:
```kotlin
if (obj.label != null) lastKnownLabel = obj.label
```

Scenario:
1. User locks on a bowl: `lockedLabel = "bowl"`, `lastKnownLabel = "bowl"`
2. Object stays in frame, detector flickers label to "potted plant" on a direct match
3. `updateFromMatch()` sets `lastKnownLabel = "potted plant"`
4. Object lost, re-acquisition starts
5. Label scoring now rewards "potted plant" candidates (wrong) and penalizes "bowl" candidates (correct)

## Fix

Change line 258 to use `lockedLabel`:

```kotlin
val labelScore = if (lockedLabel != null && candidate.label == lockedLabel) 1f else 0f
```

## Impact

With the 20% label weight, a correct "bowl" candidate loses 0.2 points while a wrong-category "potted plant" gains 0.2 — a 0.4 swing that can cause re-acquisition to pick the wrong object.

## Tests to add

Add a test that:
1. Locks with label "bowl"
2. Simulates a direct match that updates `lastKnownLabel` to "potted plant"
3. Verifies re-acquisition still prefers candidates labeled "bowl"

tuky191:
Fixed in PR #10

Issue #1, opened 2026-04-15
Bug: Double zoom calculation causes 2x zoom speed


tuky191:
## Problem

`ZoomController.calculateZoom()` is called **twice per frame** in `CameraViewModel`. Since `ZoomController` is stateful (it mutates `currentZoom` internally), each call advances the zoom by `zoomSpeed`. Two calls = double the configured speed.

## Location

`app/src/main/java/com/haptictrack/ui/CameraViewModel.kt` lines 57 and 75.

## What happens

```kotlin
// First call — line 57: applies zoom to camera hardware
val targetZoom = zoomController.calculateZoom(
    lockedObject.boundingBox,
    cameraManager.getMinZoom(),
    cameraManager.getMaxZoom()
)
cameraManager.setZoomRatio(targetZoom)

// Second call — line 74: inside _uiState.update block
currentZoomRatio = lockedObject?.let {
    zoomController.calculateZoom(   // ← mutates currentZoom AGAIN
        it.boundingBox,
        cameraManager.getMinZoom(),
        cameraManager.getMaxZoom()
    )
} ?: current.currentZoomRatio
```

The second call at line 74 is only used to populate `currentZoomRatio` in the UI state, but it advances the internal `currentZoom` a second time.

## Fix

Store the result from the first `calculateZoom()` call in a local variable and reuse it in the `_uiState.update` block:

```kotlin
val targetZoom = if (lockedObject != null) {
    zoomController.calculateZoom(
        lockedObject.boundingBox,
        cameraManager.getMinZoom(),
        cameraManager.getMaxZoom()
    ).also { cameraManager.setZoomRatio(it) }
} else null

_uiState.update { current ->
    current.copy(
        // ...
        currentZoomRatio = targetZoom ?: current.currentZoomRatio
    )
}
```

## Impact

Zoom-in and zoom-out happen at double the configured `zoomSpeed` (0.05 → effective 0.10 per frame). Makes auto-zoom feel jittery/aggressive.

tuky191:
Fixed in PR #10

PR #26, opened 2026-04-16
Person attribute recognition with face-based gender

### LINKED ISSUE #24: Feature: Person attribute recognition (gender, age, clothing)
## Problem

EfficientDet-Lite2 detects all people as generic "person" — cannot distinguish between individuals. When tracking loses a person in a crowd, re-acquisition picks the wrong one because all candidates have the same label and similar embeddings.

## Solution: Two-tier person attribute recognition

Run lightweight attribute classifiers on person crops from EfficientDet-Lite2 to extract discriminative features.

### Tier 1: Intel Crossroad-0230 (~3 MB TFLite)
- **Source**: [PINTO Model Zoo #124](https://github.com/PINTO0309/PINTO_model_zoo/tree/main/124_person-attributes-recognition-crossroad-0230) — TFLite already converted
- **Input**: 160×80 BGR person crop
- **Params**: 0.735M / 0.174 GFLOPs
- **8 binary attributes**: is_male, has_bag, has_backpack, has_hat, has_longsleeves, has_longpants, has_longhair, has_coat_jacket
- **Bonus**: outputs color sampling coordinates for upper/lower body clothing color
- **F1**: is_male 0.91, has_longpants 0.83, has_longhair 0.83, has_backpack 0.77 (has_longsleeves weak at 0.21)
- **Speed**: ~3-5ms on Snapdragon 8 Gen 2

### Tier 2: PP-LCNet pedestrian attributes (6.7 MB, needs conversion)
- **Source**: PaddlePaddle PULC `PPLCNet_x1_0_pedestrian_attribute`
- **26 attributes**: gender, age group (under 18 / 18-60 / over 60), orientation, upper/lower clothing type and pattern, boots, bag types, holding object
- **Export**: paddle model → paddle2onnx → ONNX (opset 14+ for HardSwish)
- **Speed**: ~5-10ms estimated on mobile

### Clothing color (zero-cost addition)
- Use Crossroad-0230's color sampling points to sample upper/lower body RGB
- Quantize to named colors via HSV thresholds
- Gives "red jacket", "blue pants" level discrimination

## Integration plan

1. Download Crossroad-0230 TFLite from PINTO model zoo → `app/src/main/assets/`
2. Create `PersonAttributeClassifier.kt` — runs on person crops, returns attribute set
3. Add `PersonAttributes` data class (gender, age estimate, clothing descriptors, accessories)
4. Wire into `ObjectTracker` — classify person crops during detection
5. Use attributes in `ReacquisitionEngine` scoring — "man with backpack and hat" vs "woman with handbag"
6. Add to debug frame annotations for verification

## Acceptance criteria
- [ ] Can distinguish man vs woman in re-acquisition
- [ ] Clothing attributes (hat, bag type, sleeve length) influence scoring
- [ ] Clothing color sampled and used for discrimination
- [ ] Debug frames show detected attributes
- [ ] Unit tests for attribute scoring integration

## Related
- Extends #23 (better classification)
- Uses existing EfficientDet-Lite2 person crops

### LINKED ISSUE #25: Improvement: Save raw + annotated debug frames for offline replay
## Problem

Debug frame capture currently saves only annotated frames (with bounding boxes, labels drawn on top). This makes it impossible to:
- Re-run embedding/segmentation on captured frames offline
- Compare model outputs against ground truth
- Recreate tracking scenarios from captured data

## Solution

For each debug capture event, save two frames:
1. **Raw frame** — original bitmap as captured by the camera, no annotations
2. **Annotated frame** — current behavior with bounding boxes, labels, tracking info drawn on top

### Naming convention
```
session_20260416_092119_car/
├── 092119_126_LOCK.png          # annotated (current)
├── 092119_126_LOCK_raw.png      # raw (new)
├── 092131_046_LOST.png
├── 092131_046_LOST_raw.png
└── session.log
```

## Acceptance criteria
- [ ] Every debug frame event saves both raw and annotated versions
- [ ] Raw frames are unmodified camera bitmaps
- [ ] Storage limit updated to account for ~2x frame count
- [ ] Existing session.log format unchanged

tuky191:
## Summary

- **Person attribute classification** using Intel Crossroad-0230 (2.8 MB): 8 body attributes (clothing, accessories, hair) + clothing color from fixed-region HSV sampling
- **Face-based gender + age** using BlazeFace (224 KB) + age-gender-retail-0013 (4.1 MB): 95.8% gender accuracy from face crops, overrides body-based gender when face is visible
- **Attributes integrated into re-acquisition scoring** at 15% weight — discriminates "man with backpack" vs "woman with handbag"
- **Raw + annotated debug frames** saved on every tracking event for offline replay
- **Front/back camera switch** button
- **Clothing color detection** with chromatic override for indoor lighting (gray-dominant scenes)

## Model budget
| Model | Size | Purpose |
|---|---|---|
| EfficientDet-Lite2 | 7.2 MB | Object detection |
| MobileNetV3 Large | 10 MB | Visual embedding |
| VitTracker | 0.7 MB | Frame-to-frame tracking |
| magic_touch | 5.9 MB | Interactive segmenter |
| Crossroad-0230 | 2.8 MB | Body attributes |
| BlazeFace Short Range | 0.2 MB | Face detection |
| age-gender-retail-0013 | 4.1 MB | Face gender + age |
| **Total** | **30.9 MB** | |

## Test plan
- [x] 146 unit tests pass (19 new for PersonAttributes + quantizeColor)
- [x] Device tested: front camera selfie — gender correctly classified as "man"
- [x] Re-acquisition uses attributes (scores 0.49-0.87 for correct person)
- [x] Non-person objects rejected correctly
- [ ] Outdoor test with multiple people at distance
- [ ] Verify gender fallback when face not visible (person facing away)

Closes #24, closes #25


tuky191 (commented):
## PR #26 Review: Person Attribute Recognition

Substantial feature — 3 new ML models, attribute scoring, face-based gender, front camera. 732 additions across 14 files. The architecture is sound but there are performance concerns and a weight calculation bug.

---

### High

**1. Score weight redistribution doesn't sum to 1.0 in all paths**

`ReacquisitionEngine.kt:306-313`: The position decay redistribution logic has inconsistent shares:

```kotlin
val effectiveSizeWeight = baseSizeWeight + redistributed * 0.10f
val effectiveLabelWeight = baseLabelWeight + redistributed * 0.05f
val effectiveAttrWeight = if (hasAttrs) baseAttrWeight + redistributed * 0.15f else 0f
val colorRedistShare = if (hasColor) 0.20f else 0f
val appearRedistShare = 1f - 0.10f - 0.05f - 0.15f - colorRedistShare
```

When `hasAttrs=true` and `hasColor=true`: redistribution shares = 0.10 + 0.05 + 0.15 + 0.20 + (1-0.10-0.05-0.15-0.20) = 0.10 + 0.05 + 0.15 + 0.20 + 0.50 = 1.0 ✓

When `hasAttrs=true` and `hasColor=false`: redistribution shares = 0.10 + 0.05 + 0.15 + 0 + (1-0.10-0.05-0.15-0) = 0.10 + 0.05 + 0.15 + 0.70 = 1.0 ✓

When `hasAttrs=false` and `hasColor=true`: `effectiveAttrWeight = 0f`, but `0.15f` is still subtracted from `appearRedistShare`. The 15% redistribution share meant for attrs goes nowhere:
- Shares used: 0.10 + 0.05 + 0 (attr=0) + 0.20 + (1-0.50) = 0.85
- 15% of `redistributed` is lost → total weights < 1.0

When `hasAttrs=false` and `hasColor=false`: same issue — `appearRedistShare = 1-0.10-0.05-0.15-0 = 0.70`, but attr gets 0. Total redistribution = 0.10+0.05+0+0+0.70 = 0.85.

The fix: `appearRedistShare` should be `1f - 0.10f - 0.05f - (if (hasAttrs) 0.15f else 0f) - colorRedistShare`.

**2. Person classifier runs on every candidate during search — 3 ML inferences per person**

`ObjectTracker.kt:232`: `personClassifier.classify()` is called for every candidate in the detector path. This runs BlazeFace (face detection) + age-gender model + Crossroad-0230 — 3 model inferences per person candidate. With 5 detections, that's potentially 15 extra inferences per frame on top of the segmenter + embedder.

The `classify()` method short-circuits for non-person labels (returns null), which helps. But for scenes with multiple people, this will blow the 33ms frame budget. Consider:
- Only classifying the top-N candidates by score (after a first pass without attributes)
- Caching attributes for stable tracking IDs across frames

**3. `classifyFace` crops the person AGAIN from the full frame**

`PersonAttributeClassifier.kt:150-157`: `classifyFace()` re-crops the person from the full bitmap, even though `classify()` already cropped and resized it (line 75). The person crop should be passed to `classifyFace()` rather than the full frame + box.

Additionally, the person crop at line 75 is resized to 80x160 (for the body model), but `classifyFace` needs the original-resolution crop for face detection. So the full-res crop is needed — but it's not preserved. This means a second `Bitmap.createBitmap` allocation for the same region. Consider caching the full-res crop and passing it through.

**4. `PersonAttributes.rawProbabilities` is a mutable `FloatArray` in a data class**

`TrackingState.kt:31`: Same issue as `embedding` — `FloatArray` is mutable and excluded from `equals`/`hashCode`. This is consistent with the existing pattern but the `equals`/`hashCode` override doesn't mention `rawProbabilities` in the comment, while it does exclude it. The `similarity()` method reads `rawProbabilities` for soft comparison — if the array is mutated externally, scoring silently breaks. The `copyOf()` in `lock()` protects the stored reference, but candidates pass theirs by reference.

---

### Medium

**5. `bitmapToByteBuffer` uses unnormalized [0, 255] values for body model**

`PersonAttributeClassifier.kt:243-247`: Body model gets `Color.red(pixel).toFloat()` (0-255 range). Face model at line 187 gets `Color.red(pixel).toFloat() / 255f` (0-1 range). This is likely correct if the models expect different input ranges, but it's worth verifying — Crossroad-0230 docs should specify if it expects [0,255] or [0,1] input.

**6. `switchCamera` doesn't clear tracking state**

`CameraManager.kt:74-85`: Switching cameras rebinds use cases but doesn't clear the current lock. The visual tracker, embedding gallery, and re-acquisition state all reference the old camera's frame geometry. After switching, the tracker will be fed frames from a different camera with different FOV/resolution. Consider calling `clearLock()` from the ViewModel's `switchCamera()`.

Let me check... `CameraViewModel.kt:151-153` just calls `cameraManager.switchCamera()` without clearing. The next frame from the new camera will have different content, and the visual tracker will likely lose confidence and trigger re-acquisition. It might work by accident, but it's fragile.

**7. No test for attribute scoring in `ReacquisitionEngine`**

The existing `ReacquisitionEngineTest` tests color histogram scoring, but there are no tests verifying that person attributes affect the re-acquisition score (the 15% weight path). The `PersonAttributesTest` only tests the `similarity()` and `quantizeColor()` functions in isolation. Need an integration test showing that a candidate with matching attributes scores higher than one without.

**8. Front camera mirror flip not handled for face detection**

Front camera produces mirrored frames. `FaceDetector` may return mirrored bounding boxes. If the preview is also mirrored (standard Android behavior), this cancels out for display. But if attribute classification uses the mirrored crop, left/right clothing regions may be swapped (e.g., `dominantRegionColor` column exclusion assumes a specific orientation). Worth testing on device.

---

### Low

**9. `quantizeColor` has a dead `else` branch**

`PersonAttributeClassifier.kt:335`: The `else -> "red"` branch can never be reached because the preceding `when` conditions cover the full 0-360 range. The `h >= 345f` in the first branch and `h < 345f` in the pink branch cover everything. Harmless but confusing.

**10. Raw + annotated frames doubles disk usage**

`DebugFrameCapture` now saves both raw and annotated PNGs per event. For a typical session with ~20 events and 1080p frames, that's ~200MB. The MAX_SESSIONS=20 limit applies to sessions, not total disk. Consider a lower MAX_SESSIONS or only saving raw for LOCK/REACQUIRE events.

**11. `PersonAttributes.similarity()` treats all 8 boolean attributes as equally important**

In the hard comparison fallback (line 72-80), gender mismatch has the same weight as long-hair mismatch. Gender is a much stronger discriminator. The soft comparison (line 51-57) uses raw probabilities which naturally weight more confident predictions higher — better, but a man misclassified at 0.6 confidence still looks similar to a woman at 0.4 confidence.

**12. APK model budget now 30.9MB**

The PR description correctly tallies this. The 3 new models add 7.1MB. This is within reason for a camera app, but worth noting — the app has grown from ~10MB to ~31MB in models across a few PRs.

---

### What's Good

- **Face-based gender override** is the right architecture — body-based gender from Crossroad-0230 is notoriously unreliable on close-ups, while BlazeFace + age-gender on face crops gives 95%+ accuracy.
- **Chromatic override for indoor lighting** in `dominantRegionColor()` — smart handling of the desaturation problem where gray dominates over the actual clothing color.
- **`quantizeColor` using Berlin & Kay basic color terms** — linguistically grounded, 11 terms is the right granularity for clothing.
- **19 unit tests** for PersonAttributes and quantizeColor — good coverage of the similarity function and color quantization.
- **Session logging** already wired to the `PersonAttributeClassifier` via `debugCapture.log()`.
- **`classify()` short-circuits on non-person labels** — no wasted inference.
- **Raw frame saving** is excellent for offline debugging and replay.

---

### Summary

The person attribute architecture is well-designed — face-based gender override, chromatic color correction, and soft probability scoring are all sound decisions. Main concerns: the weight redistribution bug (#1) causes scores to not sum to 1.0 when attributes are absent, the 3-model inference per candidate may blow frame budget (#2), and the double person crop allocation (#3) wastes memory. Also need integration tests for attribute scoring in the reacquisition engine, and `switchCamera` should clear tracking state.

tuky191 (commented):
## Follow-up Review: Post-fix commit a674c5f

All high-severity and most medium-severity issues addressed. Clean, well-structured fix.

### Verified Fixed

| # | Issue | Status |
|---|-------|--------|
| 1 | Weight redistribution doesn't sum to 1.0 | ✅ Redistribution shares now computed dynamically: `attrRedistShare` and `colorRedistShare` are 0 when their signals are absent, `appearRedistShare = 1 - size - label - attr - color` adapts correctly. Traced all 4 combinations — all sum to 1.0. |
| 2 | 3 ML inferences per person candidate | ✅ Two-pass approach: first pass computes embeddings+histograms for all candidates, second pass classifies person attributes for only the top-2 candidates by embedding similarity. Cuts worst-case from 15 to 6 inferences. |
| 3 | Double person crop in `classifyFace` | ✅ `cropFullRes()` creates one full-res crop, shared between body model (downscaled to 80x160) and `classifyFaceFromCrop()`. `classifyFaceFromCrop` no longer re-crops from the full frame. Both bitmaps properly recycled in `finally`. |
| 6 | `switchCamera` doesn't clear tracking | ✅ `clearTracking()` called before `cameraManager.switchCamera()` |
| 7 | No integration test for attribute scoring | ✅ Two new tests: `matching person attributes boost score over mismatching` and `no person attributes does not crash and redistributes weight`. `obj()` helper updated to accept `personAttributes`. |
| 9 | Dead `else` branch in `quantizeColor` | ✅ Last branch simplified to `else -> "pink"` with comment `// 290-345` |
| 10 | Raw + annotated doubles disk usage | ✅ `MAX_SESSIONS` reduced from 20 to 10 |
| — | `TrackedObject.equals/hashCode` comment | ✅ Updated to mention all 3 excluded fields: embedding, colorHistogram, personAttributes |

### Not changed (acceptable)

| # | Issue | Status |
|---|-------|--------|
| 4 | `rawProbabilities` mutable in data class | Consistent with existing `embedding` pattern — acceptable |
| 5 | Body model input normalization [0-255] vs [0-1] | Not changed — likely correct per model spec, needs device verification |
| 8 | Front camera mirror flip for clothing regions | Not changed — needs device testing |
| 11 | All boolean attributes weighted equally in hard fallback | Not changed — soft comparison via `rawProbabilities` is the primary path |

### Design note on the two-pass optimization

The top-2-by-embedding-similarity approach is smart — it ensures person classification only runs for the most visually relevant candidates. One edge case: if `reacquisition.bestGallerySimilarity()` is called before `processFrame()`, the gallery is from the locked object, not the current candidate set. This is correct — we want to classify the candidates most similar to the locked person, which is exactly what the sort does.

### Summary

All high-severity issues resolved. Weight redistribution is now correct across all signal combinations. Person classification cost cut from O(n) to O(2) per frame. Single-crop-shared-across-models eliminates the duplicate allocation. Tests pass (148 total). Good to merge.

Issue #30, opened 2026-04-16
Reproducible re-acquisition testing + scoring simplification


tuky191:
## Problem

The re-acquisition scoring system has grown complex (6 weighted signals, redistribution logic, decay, overrides) and is hard to reason about. Each fix risks regressions elsewhere. Testing relies on ad-hoc device tests — there's no way to replay a scenario deterministically.

Symptoms from recent work (#28, #29):
- Chair-hopping: locks on one chair, loses it, reacquires a different chair, repeats
- Cross-category leakage: "person" reacquired when locked label was "chair" (high color similarity compensated for wrong label)
- Weight tuning is whack-a-mole — fixing one case breaks another

## Root cause: weighted average scoring

Production trackers (DeepSORT, ByteTrack, BoT-SORT) use **cascade/gating**, not weighted averages. Our 6-signal weighted score lets a strong signal in one dimension override a bad match in another — which is exactly the chair→person failure mode.

| Our approach | Production trackers (DeepSORT) |
|---|---|
| 6 signals × weights → single score | 2-3 signals in hard cascade |
| Strong color can override wrong label | Wrong label = immediate reject |
| All signals compete in one pass | Each stage filters, then passes down |

DeepSORT pattern: gate on appearance (cosine sim > threshold), then rank by position. A person never passes the appearance gate for a chair — no amount of color similarity rescues it.

### Phase 1: Scenario recorder + replay harness

Build infrastructure to turn device test failures into reproducible unit tests.

**Scenario recorder** (`ScenarioRecorder.kt`):
- Toggles on at lock time, records every `processFrame` input
- Serializes per frame: `List<TrackedObject>` (id, label, confidence, box, embedding, colorHistogram, personAttributes)
- Also records: locked object state, gallery embeddings, lock-time metadata
- Writes JSON to debug_frames directory alongside PNGs
- Stop on clear/timeout

**Replay harness** (`ScenarioReplayTest.kt`):
- Reads scenario JSON from `src/test/resources/scenarios/`
- Calls `engine.lock()` with saved initial state
- Loops through frames calling `engine.processFrame(detections)`
- Asserts events at specific frames (LOST at frame N, REACQUIRE at frame M, correct target ID)
- Runs in Robolectric like existing tests — no device needed

**Format**: JSON, one file per scenario. ~2MB for a 450-frame scenario with 5 detections/frame including embeddings.

### Phase 2: Refactor scoring to cascade

[...]

PR #31, opened 2026-04-17
Cascade scoring + scenario replay testing (#30)

tuky191:
## Summary

- **Phase 1**: `ScenarioRecorder` captures every `processFrame()` input as JSON during live tracking sessions. `ScenarioReplayTest` replays them through a fresh `ReacquisitionEngine` for deterministic regression testing. Any device bug → pull scenario.json → add replay test → never regresses.
- **Phase 2**: Replace 6-signal weighted average in `scoreCandidate()` with DeepSORT-style cascade gates. Wrong-label candidates are hard-rejected at the label gate — only strong embedding (>0.7) overrides. Survivors ranked by embedding similarity (50%) with position/size/color/attrs as tiebreakers.

## What changed

| File | Change |
|------|--------|
| `ScenarioRecorder.kt` | NEW — JSON capture of lock state + per-frame detections with embeddings |
| `ScenarioReplayTest.kt` | NEW — replay harness + 7 tests (3 synthetic, 4 real scenarios) |
| `ReacquisitionEngine.kt` | Rewrote `scoreCandidate()`: label gate + simplified ranking |
| `ReacquisitionEngineTest.kt` | Updated 7 tests, added 7 cascade-specific tests (74 total) |
| `ObjectTracker.kt` | Wires ScenarioRecorder into lock/frame/clear lifecycle |
| `DebugFrameCapture.kt` | Exposed `sessionDir` for ScenarioRecorder |
| `CLAUDE.md` | Full docs: cascade scoring, replay workflow, source layout |
| `scenarios/*.json` | 2 captured fixtures: cup reacquisition, mouse cascade |

## Device test results

**Mouse session** (cascade): 8 re-acquisitions, all "mouse", zero cross-category leakage. Keyboards, TVs, persons, scissors all rejected by label gate. Hop counter stayed at 0-1 (same-object detection working).

**Cup session** (cascade): 4 re-acquisitions, all "cup". Correctly rejected bed, vase, book, toothbrush, handbag. Hit hop limit after 3 hops — hop counter tuning is a separate follow-up.

**Before cascade**: chair sessions hopped between different chairs and eventually locked a person. Potted plant sessions locked bowls. The cascade eliminates all cross-category leakage.

## Test plan

- [x] 172 unit tests pass (74 ReacquisitionEngine + 7 ScenarioReplay + rest)
- [x] cup_reacquisition.json replay passes
- [x] mouse_cascade_reacquisition.json replay passes
- [x] Device tested: mouse tracking across angles/distances
- [x] Device tested: cup tracking with camera movement
- [ ] Capture more scenarios (chair, potted plant, bottle) for broader coverage


tuky191 (commented):
## Review: PR #31 — Cascade scoring + scenario replay testing

Good work. The cascade gates are a clean improvement over the weighted average — the chair→person leakage fix is exactly right. ScenarioRecorder + replay harness is well-designed infrastructure. 172 tests pass.

### Issues

#### 1. Thread safety: ScenarioRecorder accessed from multiple threads (High)

`ScenarioRecorder` has no synchronization, but is called from different threads:
- **UI thread**: `lockOnObject()` → `start()`, `clearLock()` → `recordEvent()`/`stop()`
- **Camera thread**: `processImage()` → `recordFrame()`/`recordEvent()`

The `recording` flag, `frames` JSONArray, `events` JSONArray, and `frameIndex` are all shared mutable state. A race between `clearLock` (UI) and `processImage` (camera) could corrupt the JSON or cause a concurrent modification.

Fix: Either mark methods `@Synchronized` or use an internal lock. The recorder does lightweight work per call, so contention won't be an issue.

#### 2. Per-frame Base64 encoding on camera thread (Medium)

`recordFrame()` calls `trackedObjectToJson()` which calls `floatArrayToBase64()` for every detection's embedding (1280 floats → ~5KB) and color histogram (26 floats). At 30fps with 5 detections/frame, that's 150 Base64 allocations/second plus JSONObject construction on the camera analysis thread.

Consider deferring serialization — store the raw `TrackedObject` list and serialize lazily in `flush()`, which runs after `stop()`. This moves the heavy work off the camera thread.

#### 3. `assertTimesOut` is too permissive (Medium)

`ScenarioReplayTest.kt:71`:
```kotlin
assertTrue("...", timeout || !reacquire)
```

This passes when there are **zero events at all** (no TIMEOUT, no REACQUIRE). A scenario with zero events shouldn't be considered a "timeout." Should be:
```kotlin
assertTrue("...", timeout && !reacquire)
```
or at minimum just `assertTrue("...", timeout)`.

#### 4. ScenarioRecorder not stopped on app shutdown (Medium)

`ObjectTracker.shutdown()` (line 446) doesn't call `scenarioRecorder.stop()`. If the app is killed/paused mid-session, the scenario JSON is never flushed to disk. Add `scenarioRecorder.stop()` to `shutdown()`.

#### 5. Stale test comments from weighted-scoring era (Low)

`ReacquisitionEngineTest.kt:1004` — comment says "should get penalty" but with cascade gates, wrong-label candidates are now either hard-rejected or pass via embedding override. The test still works correctly (wrong candidate passes via sim=1.0 override, loses the +0.05 label bonus) but the comment is misleading.

Same issue at line 1032 — "table" candidate comment implies it gets a penalty but it actually passes via embedding override.

#### 6. Fragile null handling in `jsonToPersonAttributes` (Low)

`ScenarioRecorder.kt:204`:
```kotlin
obj.optString("upperColor", null).takeIf { it != "null" }
```

When JSON contains `"upperColor": null`, `optString` returns the string `"null"`, and the `takeIf` catches it. This works but is fragile — if someone stores a color literally named "null", it breaks. Use the same pattern as the other nullable fields:
```kotlin
if (!obj.isNull("upperColor")) obj.getString("upperColor") else null
```

#### 7. O(n²) JSONArray clearing in `start()` (Low)

`ScenarioRecorder.kt:65-66`:
```kotlin
frames.let { while (it.length() > 0) it.remove(0) }
```

`JSONArray.remove(0)` shifts all elements on each call — O(n²) for n elements. Since `start()` should only be called when not recording (and the arrays should be empty), this is unlikely to matter in practice. But it's cleaner to just replace with new instances:
```kotlin
private var frames = JSONArray()
// in start():
frames = JSONArray()
events = JSONArray()
```

- **VT frames not recorded**: When the visual tracker is active (ObjectTracker lines 143-226), `recordFrame()` is never called. Replayed scenarios only capture the detector path. This is fine for testing `ReacquisitionEngine`, but worth documenting — the scenario doesn't represent the full session lifecycle.

- **Weight math is correct**: Verified all weight combinations sum to 1.0 (before label bonus). The `unused` redistribution to embedding is clean.

- **Cascade gates logic is sound**: Gate A (geometry) → Gate B (label) → Ranking. Embedding override at each gate is correctly checked. The `strongVisualMatch` variable is computed once and reused across all gates.

### Verdict

Issues #1 (thread safety) and #4 (shutdown) should be fixed before merge — both can lose data silently. The rest are improvements that can be addressed now or later.

tuky191 (commented):
## Follow-up review: commit 6036a7c

All 7 issues addressed:

| # | Issue | Status |
|---|-------|--------|
| 1 | Thread safety | `@Synchronized` on all 4 public methods |
| 2 | Per-frame Base64 | Deferred to `flush()` via raw `FrameData`/`EventData` |
| 3 | `assertTimesOut` permissive | Now requires actual TIMEOUT event |
| 4 | Recorder not stopped on shutdown | Added `scenarioRecorder.stop()` to `shutdown()` |
| 5 | Stale test comments | Updated to reflect cascade gate behavior |
| 6 | Fragile null handling | Uses `obj.isNull()` pattern consistently |
| 7 | O(n²) JSONArray clearing | Replaced with fresh `MutableList` assignment |

Good to merge.

Issue #34, opened 2026-04-17
Person identity: face embedding + body re-ID for person-vs-person discrimination


tuky191:
## Problem

Current re-acquisition can't reliably distinguish between two people. MobileNetV3 embeddings are trained for general visual similarity, not within-category identity. Two different people in similar clothing get sim ~0.5-0.7, same as the same person from a different angle.

The cascade label gate prevents cross-category leakage (person locked as chair), but within "person" we have weak discrimination. Current signals:
- MobileNetV3 embedding: not designed for person re-ID
- Color histogram: fails when people wear similar colors
- Person attributes (Crossroad-0230): coarse — gender, clothing type, accessories. Two similar men are indistinguishable.

## Proposed solution: progressive identity building

Build identity signals incrementally during tracking, use the strongest available for re-acquisition:

### Signal hierarchy (strongest → weakest)

1. **Face embedding** (MobileFaceNet or similar, ~1-5MB) — near-certain identity when face visible. Added to lock attributes when face first appears during confirmed tracking.
2. **Body re-ID embedding** (OSNet-AIN or similar, ~2MB) — full-body identity encoding that captures body type, proportions, gait, clothing. Available from first lock.
3. **Person attributes** (Crossroad-0230, existing) — gender, clothing, accessories. Coarse but always available.
4. **Body proportions** (MediaPipe Pose landmarks or bounding box ratios) — shoulder/hip ratio, height/width as proxy for body type (fit/normal/obese).
5. **Clothing color** (existing histogram) — shirt/pants color.
6. **Generic embedding** (MobileNetV3, existing) — general visual similarity.

### Lock lifecycle

```
Initial lock (any angle):
  → body re-ID embedding (OSNet)
  → person attributes (Crossroad-0230)  
  → clothing colors (histogram)
  → generic embedding (MobileNetV3)

During confirmed tracking, face appears:
  → face embedding computed and stored
  → strongest identity signal now available

Re-acquisition:
  → cascade through signals by strength
  → face visible + face embedding stored → high-confidence match
  → no face → fall back to body re-ID + attributes + color
  → nothing matches → clean loss (don't hop to another person)
```

### Key principle

**Clean cutover on loss.** If the best available signals don't match above threshold, accept the loss rather than locking onto a different person. Better to lose tracking than to track the wrong person.

## Research needed

[...]

PR #36, opened 2026-04-17
Person identity: face embedding + body re-ID (#34)

tuky191:
## Summary

- **MobileFaceNet** (5.0 MB, 192-dim): face identity embedding. Computed at lock time + progressively when face appears during tracking.
- **OSNet x1.0** (4.3 MB fp16, 512-dim): person re-ID embedding trained on Market-1501. Available from first lock on any person.
- Progressive identity building: lock from behind → re-ID + attributes, face appears → face embedding added
- Cascade ranking uses strongest available signal: face (45%) > re-ID (40%) > generic embedding (50%) > position/size

## New files

| File | What |
|------|------|
| `FaceEmbedder.kt` | BlazeFace face detection + MobileFaceNet embedding |
| `PersonReIdEmbedder.kt` | OSNet body crop → 512-dim re-ID embedding |
| `mobilefacenet.tflite` | 5.0 MB, [2,112,112,3] → [2,192] |
| `osnet_x1_0_market.tflite` | 4.3 MB, [1,256,128,3] → [1,512] |

## Device test results

- Both models load and run successfully on device
- Face embedding computed at lock time and during search (norm=1.00 after L2)
- Re-ID embedding computed for all person candidates
- Cascade correctly rejects non-person labels (suitcase, tv, keyboard, potted plant)
- Re-acquisition finds the correct person with re-ID + face + attrs scoring

## Test plan

- [x] 166 unit tests pass
- [x] Build succeeds (47 MB total model size)
- [x] Device test: lock on person, face embedding captured at lock
- [x] Device test: re-acquisition uses re-ID + face in scoring
- [ ] Test with two people in frame — verify it tracks the right one

### DIFF
--- a/app/src/main/java/com/haptictrack/tracking/EmbeddingUtils.kt
+++ b/app/src/main/java/com/haptictrack/tracking/EmbeddingUtils.kt
@@ -155,6 +155,19 @@ fun computeIou(a: RectF, b: RectF): Float {
     return if (unionArea > 0f) interArea / unionArea else 0f
 }
 
+/** L2 norm of a float array. */
+fun l2Norm(arr: FloatArray): Float {
+    var sum = 0f
+    for (v in arr) sum += v * v
+    return kotlin.math.sqrt(sum)
+}
+
+/** L2-normalize a float array in place. */
+fun l2Normalize(arr: FloatArray) {
+    val n = l2Norm(arr)
+    if (n > 1e-6f) for (i in arr.indices) arr[i] /= n
+}
+
 /**
  * Histogram correlation — measures how similar two color distributions are.
  * Returns a value in [-1, 1] where 1 = identical distributions.
--- a/app/src/main/java/com/haptictrack/tracking/FaceEmbedder.kt
+++ b/app/src/main/java/com/haptictrack/tracking/FaceEmbedder.kt
@@ -87,7 +87,7 @@ class FaceEmbedder(context: Context, sharedFaceDetector: FaceDetector? = null) {
         if (personCrop.width < 30 || personCrop.height < 30) return null
         try {
             val mpImage = BitmapImageBuilder(personCrop).build()
-            val faces = faceDetector.detect(mpImage)
+            val faces = synchronized(faceDetector) { faceDetector.detect(mpImage) }
             if (faces.detections().isEmpty()) return null
 
             // Use the largest face
--- a/app/src/main/java/com/haptictrack/tracking/FaceEmbedder.kt
+++ b/app/src/main/java/com/haptictrack/tracking/FaceEmbedder.kt
@@ -23,7 +23,7 @@ import java.nio.ByteOrder
  * Used for person-vs-person discrimination when a face is visible.
  * The embedding is added to the lock attributes progressively — not required at lock time.
  */
-class FaceEmbedder(context: Context) {
+class FaceEmbedder(context: Context, sharedFaceDetector: FaceDetector? = null) {
 
     companion object {
         private const val TAG = "FaceEmbed"
@@ -36,6 +36,7 @@ class FaceEmbedder(context: Context) {
 
     private val interpreter: Interpreter
     private val faceDetector: FaceDetector
+    private val ownsFaceDetector: Boolean
 
     // Pre-allocated buffers — this MobileFaceNet variant has fixed batch=2
     private val inputBuffer: ByteBuffer = ByteBuffer.allocateDirect(4 * 2 * INPUT_SIZE * INPUT_SIZE * 3).apply {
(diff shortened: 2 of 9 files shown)


tuky191 (commented):
## Review: PR #36 — Person identity: face embedding + body re-ID (#34)

Good addition — MobileFaceNet + OSNet gives the pipeline real person-vs-person discrimination instead of relying on generic MobileNetV3 + clothing color. Progressive face building is a nice design. Weight math checks out across all tier combinations.

### Issues

#### 1. Coordinate mismatch: `vtBox` vs `rawBox` for progressive face embedding (High)

`ObjectTracker.kt:187`:
```kotlin
val faceEmb = faceEmbedder.embedFace(bitmap, vtBox)
```

`vtBox` is in **screen space** (unmapped), but `bitmap` is the **rotated** image. Two lines above, the gallery embedding correctly uses `rawBox` (rotated-image coords):
```kotlin
val emb = appearanceEmbedder.embedWithFallback(bitmap, rawBox)  // line 180 ✓
```

With deviceRot=0 this is invisible, but in landscape/upside-down the face crop will be from the wrong region — likely producing a garbage embedding that poisons identity matching.

Fix: `faceEmbedder.embedFace(bitmap, rawBox)`

#### 2. Pre-allocated buffers are not thread-safe (High)

`FaceEmbedder.inputBuffer`/`outputArray` and `PersonReIdEmbedder.inputBuffer`/`outputArray` are mutable instance fields reused across calls. Both embedders are called from:
- **UI thread**: `lockOnObject()` → `faceEmbedder.embedFace()`, `personReId.embed()` (inside `synchronized(lastFrameLock)`)
- **Camera thread**: `processImage()` → VT path (line 187) and detector path (line 285)

The camera thread does NOT hold `lastFrameLock` when calling these embedders. If a user taps to lock while the camera thread is in the VT confirmed-frame path, both threads write to the same `inputBuffer` simultaneously → corrupted inference.

Options:
- Allocate fresh buffers per call (simplest, slight GC pressure)
- Synchronize the `embed*` methods on `this`
- Route all embedder calls through `synchronized(lastFrameLock)` in ObjectTracker

#### 3. Duplicate BlazeFace detector instances (Medium)

`FaceEmbedder` loads its own `FaceDetector` from `blaze_face_short_range.tflite`, while `PersonAttributeClassifier` already loads the same model. That's two MediaPipe detector instances for the same model in memory. Since both are only used for person crops, consider sharing one `FaceDetector` instance (e.g. injected via constructor, or a shared helper).

#### 4. Duplicate `l2Normalize`/`norm` (Medium — DRY)

Both `FaceEmbedder` (lines 137-147) and `PersonReIdEmbedder` (lines 117-127) have identical private `l2Normalize` and `norm` methods. These belong in `EmbeddingUtils.kt` alongside `cosineSimilarity` and `bestGallerySimilarity`.

#### 5. No tests for face/re-ID scoring tiers (Medium)

166 tests pass, but all existing `scoreCandidate` tests use the generic embedding path. The two new ranking tiers (face-primary at 45%, re-ID-primary at 40%) have zero test coverage. At minimum, add tests for:
- Face tier: candidate with matching face scores higher than candidate with mismatching face
- Re-ID tier: candidate with matching re-ID scores higher than mismatching
- Tier precedence: face-available candidate uses face weights, not generic weights
- Weight redistribution: verify all tiers sum to 1.0 when signals are missing

#### 6. ScenarioRecorder doesn't serialize new fields (Medium)

`trackedObjectToJson` and `jsonToTrackedObject` don't handle `reIdEmbedding` or `faceEmbedding`. Replayed scenarios will always hit the generic embedding path, never the face or re-ID tiers. Future scenario captures won't reproduce person-identity bugs.

#### 7. INVARIANT comment stale (Low)

`TrackingState.kt:121-123` — the comment says "embedding, colorHistogram, and personAttributes are excluded" but now `reIdEmbedding` and `faceEmbedding` are also excluded. Update to list all 5 excluded fields.

- **Weight math is correct** for all 4 tiers. Face tier sums to 1.0 in all signal combinations, re-ID tier likewise.
- **Progressive face gate is well-designed**: `lockedFaceEmbedding == null && lockedReIdEmbedding != null` ensures only persons get progressive face capture, and it stops after the first successful face.
- **`addFaceEmbedding` ignores subsequent calls** (line 98: `if (lockedFaceEmbedding == null)`) — good, first face wins, no churn.
- **Model size impact**: +9.3 MB (MobileFaceNet 5.0 MB + OSNet 4.3 MB), total ~47 MB. Reasonable for the identity gain.

### Verdict

Fix #1 (coordinate bug) and #2 (thread safety) before merge — both corrupt inference results silently. #3-6 are improvements worth doing in this PR or a fast follow-up.

tuky191 (commented):
## Follow-up review: commit 7db14fe

All 7 issues addressed:

| # | Issue | Status |
|---|-------|--------|
| 1 | Coordinate bug (vtBox→rawBox) | Fixed — now uses `rawBox`, comment added |
| 2 | Thread safety (pre-allocated buffers) | `@Synchronized` on `embedFaceFromCrop` and `embedFromCrop` |
| 3 | Duplicate BlazeFace | Shared via `sharedFaceDetector` param + `ownsFaceDetector` flag |
| 4 | DRY (l2Normalize/norm) | Moved to `EmbeddingUtils.kt` as `l2Normalize`/`l2Norm`, removed from both embedders |
| 5 | No tests for new tiers | 4 new tests: face matching, re-ID matching, face precedence over re-ID, generic fallback |
| 6 | ScenarioRecorder missing fields | `reIdEmbedding`/`faceEmbedding` serialized + deserialized with `obj.has()` for backward compat |
| 7 | Stale INVARIANT comment | Updated to list all 5 excluded fields |

**Minor note (not blocking):** The shared `faceDetector` is now called from `FaceEmbedder.embedFaceFromCrop` (synchronized) and `PersonAttributeClassifier.classifyFaceFromCrop` (not synchronized). If both hit the same detector concurrently from different threads, MediaPipe's internal thread safety determines the outcome. Low risk in practice since both are called sequentially within a single frame's processing, but if you want belt-and-suspenders, add `@Synchronized` to `classifyFaceFromCrop` too — or synchronize on the shared detector instance.

Good to merge.

Issue #14, opened 2026-04-15
Feature: Manual camera controls (zoom, recording speed, photo/video modes)


tuky191:
## Problem

All camera parameters are currently automatic with no user control. The app needs manual overrides for key camera functions to be useful as a real filming tool.

### 1. Manual zoom (in/out)
- On-screen zoom slider or +/- buttons
- Override auto-zoom when the user wants a specific framing
- Should respect the optical zoom limit (auto-detected, currently 3.6x on Xiaomi 13 Pro)
- Auto-zoom resumes when user releases manual control (or toggle to disable auto-zoom)

### 2. Recording speed
- Support for slow-motion and timelapse capture
- Speed options: 0.25x, 0.5x, 1x, 2x, 4x (or similar)
- CameraX VideoCapture supports frame rate control via `Quality` selector
- Slow-mo requires checking device capability (`CameraInfo` supported qualities)

### 3. Photo/video mode switch
- **Video mode** (current): continuous recording with start/stop
- **Photo interval mode**: capture still photos at a configurable interval (e.g., every 1s, 2s, 5s) while tracking
  - Useful for timelapses, documentation, or situations where video is overkill
  - Should use `ImageCapture` use case from CameraX
  - Photos saved to MediaStore like videos currently are

## UI Considerations

- Controls should be minimal and non-intrusive — this is a hands-free app
- Could use a bottom drawer or swipe-up panel for settings
- Zoom could be a vertical slider on the side (like native camera apps)
- Mode switch: simple toggle icon between video/photo near the record button
- Speed selector: small chip/pill near the record button showing current speed

## Implementation Notes

- `CameraManager` needs to expose `ImageCapture` use case alongside `VideoCapture`
- Binding both `ImageCapture` and `VideoCapture` simultaneously may not be supported on all devices — need to check `ProcessCameraProvider` capabilities
- Auto-zoom interaction: manual zoom should temporarily disable auto-zoom, with a "reset to auto" gesture or timeout
- Recording speed may require switching between `Quality.HIGHEST` and device-specific frame rates

## Files likely affected

- `app/src/main/java/com/haptictrack/camera/CameraManager.kt` — add ImageCapture, frame rate control
- `app/src/main/java/com/haptictrack/camera/RecordingManager.kt` — speed control, photo interval mode
- `app/src/main/java/com/haptictrack/ui/CameraScreen.kt` — new UI controls
[...]

PR #37, opened 2026-04-17
Manual camera controls: pinch zoom, 4K recording, stealth mode (#14)

tuky191:
## Summary

- **Pinch-to-zoom** with 2s manual override that pauses auto-zoom, then resumes
- **Volume-down 3-stage cycle**: lock center object → start recording → stop + clear (fully hands-free)
- **4K video recording** via dynamic 2/3-stream switching — drops ImageAnalysis during recording, uses PreviewView.getBitmap() for tracking instead
- **Stealth mode**: black screen overlay, tracking + recording continue via hidden preview
- **Zoom level indicator**: always visible when tracking, fades after pinch in idle
- **Capture mode pill** (VIDEO/PHOTO) with swipe gesture
- Faster auto-zoom (0.05 → 0.10 per frame)
- Camera rebind resilience: prepareForRebind() forces search mode for instant recovery

### 4K recording approach

Camera hardware limits 3 concurrent streams to 720p. When recording starts, ImageAnalysis is unbound (2 streams → 4K video). Analysis frames come from `PreviewView.getBitmap()` downscaled to ~640px width (~30ms/frame). On recording stop, 3-stream mode resumes via Finalize callback.

| Mode | Streams | Video | Analysis |
|------|---------|-------|----------|
| Tracking | preview + analysis + video | 720p | ImageAnalysis |
| Recording | preview + video | **4K** | getBitmap() |
| Stealth | preview + video (hidden) | **4K** | getBitmap() |

### Tested on Xiaomi 13 Pro
- 3840x2160 @ 30fps confirmed during recording
- Tracking + re-acquisition verified via getBitmap path (sim=0.828-0.949)
- Stealth mode: lock → stealth → record → stop → exit all working

## Test plan

- [x] Pinch-to-zoom works in idle and locked states
- [x] Volume-down cycles: lock → record → stop+clear
- [x] Zoom indicator visible during tracking, fades in idle
- [x] 4K video output confirmed via ffprobe
- [x] Tracking continues during recording via getBitmap
- [x] Stealth mode: black screen, tracking works, record works
- [x] Mode switch recovery: object reacquired within 1-2 frames
- [x] All 175 unit tests pass
- [ ] Test with person tracking (face/re-ID during getBitmap mode)
- [ ] Test stealth mode with phone orientation changes

### DIFF
--- a/app/src/main/java/com/haptictrack/MainActivity.kt
+++ b/app/src/main/java/com/haptictrack/MainActivity.kt
@@ -1,20 +1,34 @@
 package com.haptictrack
 
 import android.os.Bundle
+import android.view.KeyEvent
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.enableEdgeToEdge
+import androidx.activity.viewModels
 import com.haptictrack.ui.CameraScreen
+import com.haptictrack.ui.CameraViewModel
 import com.haptictrack.ui.theme.HapticTrackTheme
 
 class MainActivity : ComponentActivity() {
+
+    private val viewModel: CameraViewModel by viewModels()
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         enableEdgeToEdge()
         setContent {
             HapticTrackTheme {
-                CameraScreen()
+                CameraScreen(viewModel)
             }
         }
+    }
+
+    override fun onKeyDown(keyCode: Int, event: KeyEvent?): Boolean {
+        if (keyCode == KeyEvent.KEYCODE_VOLUME_DOWN) {
+            viewModel.onVolumeDown()
+            return true
+        }
+        return super.onKeyDown(keyCode, event)
     }
 }
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -48,6 +48,13 @@ class CameraManager(private val context: Context) {
     var isFrontCamera: Boolean = false
         private set
 
+    /** When true, ImageAnalysis is unbound to free a stream for higher video resolution. */
+    private var recordingMode: Boolean = false
+
+    /** Whether ImageAnalysis is currently active (false during recording). */
+    var analysisActive: Boolean = true
+        private set
+
     val preview = Preview.Builder().build()
 
     val imageAnalysis: ImageAnalysis = ImageAnalysis.Builder()
@@ -61,17 +68,22 @@ class CameraManager(private val context: Context) {
         )
         .build()
 
-    private val recorder = Recorder.Builder()
-        .setQualitySelector(
-            QualitySelector.fromOrderedList(
-                listOf(Quality.UHD, Quality.FHD, Quality.HD),
-                FallbackStrategy.higherQualityOrLowerThan(Quality.FHD)
-            )
-        )
-        .setExecutor(Executors.newSingleThreadExecutor())
-        .build()
+    private val videoExecutor = Executors.newSingleThreadExecutor()
+    var videoCapture = createVideoCapture()
+        private set
 
-    val videoCapture = VideoCapture.withOutput(recorder)
+    private fun createVideoCapture(): VideoCapture<Recorder> {
+        val recorder = Recorder.Builder()
+            .setQualitySelector(
+                QualitySelector.fromOrderedList(
+                    listOf(Quality.UHD, Quality.FHD, Quality.HD),
+                    FallbackStrategy.higherQualityOrLowerThan(Quality.FHD)
+                )
+            )
+            .setExecutor(videoExecutor)
+            .build()
+        return VideoCapture.withOutput(recorder)
+    }
 
     init {
         opticalZoomMax = detectOpticalZoomMax()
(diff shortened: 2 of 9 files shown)


tuky191 (commented):
## Review: PR #37 — Manual camera controls: pinch zoom, 4K recording, stealth mode (#14)

Big PR (564 additions, 9 files). The 2/3-stream switching for 4K is a smart approach, and the volume-down 3-stage cycle is well-thought-out for hands-free use. 175 tests pass.

### Issues

#### 1. `deviceRotationProvider` swap in `processBitmap` is a race condition (High)

`ObjectTracker.processBitmap()` saves the current `deviceRotationProvider`, swaps it to `{ 0 }`, runs processing, then restores it. But `processBitmap` runs on `Dispatchers.Default`, while `ImageAnalysis.Analyzer` runs on CameraX's analysis executor.

During the transition window (between `enterRecordingMode()` and CameraX actually unbinding analysis), both paths could be active simultaneously. One thread sees `{ 0 }`, the other sees the real provider, or worse — the restore clobbers a concurrent read.

Fix: Instead of swapping a shared mutable field, pass the rotation provider (or the rotation value) as a parameter to `processBitmapInternal`:

```kotlin
fun processBitmap(bitmap: Bitmap) {
    processBitmapInternal(bitmap, deviceRotation = 0, imageProxy = null)
}

private fun processImage(imageProxy: ImageProxy) {
    ...
    processBitmapInternal(bitmap, deviceRotation = deviceRotationProvider?.invoke() ?: 0, imageProxy = imageProxy)
}
```

Then `processBitmapInternal` uses the passed-in rotation instead of calling `deviceRotationProvider`.

#### 2. Exiting stealth while recording destroys the recording (High)

`toggleStealthMode()` calls `cameraManager.exitRecordingMode()` unconditionally when exiting stealth. `exitRecordingMode()` → `bindUseCases()` → `provider.unbindAll()` — this kills any active `VideoCapture`. The STEALTH/EXIT button is always visible, so a user could:

1. Enter stealth → start recording → exit stealth → recording silently killed

The `toggleStealthMode` exit path should check `_uiState.value.isRecording` and either block the exit or stop the recording first.

#### 3. `toggleRecording` captures `inStealth` at call time, stale in Finalize callback (Medium)

```kotlin
val inStealth = _uiState.value.stealthMode  // captured once
...
is VideoRecordEvent.Finalize -> {
    if (!inStealth) {  // uses stale value
        cameraManager.exitRecordingMode()
    }
}
```

If the user exits stealth while recording is active, `inStealth` is still `true` from when recording started. The Finalize callback skips `exitRecordingMode()`, leaving the camera in 2-stream mode with no ImageAnalysis.

Fix: Read `_uiState.value.stealthMode` inside the callback instead of capturing it.

#### 4. `needsRotation` parameter is unused (Low)

`processBitmapInternal(bitmap, needsRotation, imageProxy)` — the `needsRotation` parameter is passed as `true`/`false` but never read in the method body. The rotation behavior is actually controlled by the `deviceRotationProvider` swap. Either remove the parameter or use it instead of the provider swap (which would also fix issue #1).

#### 5. `onPinchEnd()` is dead code (Low)

`CameraViewModel.onPinchEnd()` is defined but never called. The `ScaleGestureDetector.onScaleEnd` calls `viewModel.hideZoomIndicator()` directly. Remove the dead method.

#### 6. RecordButton color has redundant branch (Low)

```kotlin
containerColor = when {
    isRecording -> HapticRed
    captureMode == CaptureMode.PHOTO -> Color.White
    else -> Color.White
}
```

Both non-recording branches return `Color.White`. Simplify to `if (isRecording) HapticRed else Color.White`.

- **`prepareForRebind` setting `framesLost = 1`** is correct — forces search mode on next processFrame, since tracking IDs reset after rebind. Good design.
- **Bitmap scaling in analysis loop** is well-handled — recycles the original before passing scaled copy, or passes through when already small enough.
- **`enterRecordingMode` is idempotent** (`if (recordingMode) return`) — safe for stealth → record transitions where both call it.
- **ZoomController manual override uses `System.currentTimeMillis()`** — untestable for expiry (the test at line 946 acknowledges this). Consider a `timeProvider` lambda for future testability, but not blocking.
- **Volume-down stage 3 (stop + clear)** — calls `toggleRecording()` then `clearTracking()`. If Finalize is async, the clear could race with the finalize callback. In practice this is likely fine since both update `_uiState` on the main thread.

### Verdict

Fix #1 (rotation provider race) and #2 (stealth exit kills recording) before merge — both corrupt tracking or lose video silently. #3 is medium risk and should also be addressed.

tuky191 (commented):
## Follow-up review: commit 9848d23

All 6 issues addressed:

| # | Issue | Status |
|---|-------|--------|
| 1 | Race condition (deviceRotationProvider) | Fixed — rotation resolved once at call site, passed as `Int` param through `processBitmapInternal` and `runDetector`. No shared mutable state. |
| 2 | Stealth exit kills recording | Fixed — `toggleStealthMode()` now stops recording before exiting stealth |
| 3 | Stale `inStealth` in Finalize | Fixed — reads `_uiState.value.stealthMode` at callback time, not capture time |
| 4 | Unused `needsRotation` param | Fixed — replaced with `deviceRotation: Int` |
| 5 | Dead `onPinchEnd()` | Removed |
| 6 | Redundant button color | Simplified |

**One concern: all 5 ZoomController manual zoom tests were deleted.** `setManualZoom`, `clamp`, `manualOverride pause`, and `reset` had valid test coverage. The only weak test was the time-based expiry one (which I noted as an observation, not a fix request). The other 4 tests are worth keeping — they verify real behavior that still exists in `ZoomController`. Recommend restoring those 4 tests.

Other than the test removal, good to merge.

PR #42, opened 2026-04-19
GPU delegate, FP16 detector, speed optimizations, stealth mode

### LINKED ISSUE #41: Improve tracking speed in stealth/recording mode
## Problem

In stealth/recording mode, frames come from `PreviewView.getBitmap()` and the entire pipeline runs sequentially on `Dispatchers.Default`. Per-frame latency is ~80-150ms, limiting effective throughput to ~7-12fps despite the 50ms loop interval.

## Current bottlenecks

| Step | Time (CPU) | Frequency |
|---|---|---|
| getBitmap() + downscale | ~30ms | Every frame |
| EfficientDet-Lite2 (MediaPipe) | ~35ms | Every frame |
| VitTracker (OpenCV DNN) | ~5-8ms | When locked |
| YOLOv8n label enrichment | ~270ms | Every 10th search frame |
| MobileNetV3 embedding | ~8ms/crop | Per candidate |
| Person attrs + re-ID + face | ~20ms | Top-2 persons |

### 1. GPU delegate for MediaPipe detector
Switch EfficientDet-Lite2 from CPU to GPU delegate (Adreno 740). Expected: ~35ms → ~15ms (~2x). Single config change.

### 2. Pipeline parallelism (VT + detector)
VisualTracker and MediaPipe detector are independent — run concurrently. Expected: ~35% wall-clock reduction when locked.

### 3. Adaptive frame skipping
When VitTracker confidence > 0.7, skip detector every 2nd-3rd frame. VT alone at ~5ms handles interim frames.

### 4. FP16 quantization for TFLite models
YOLOv8n and MobileNetV3 on Adreno 740 GPU. FP16 native, no accuracy loss. INT8 not safe for embedder.

### 5. Convert VitTracker ONNX → TFLite + GPU
Eliminates OpenCV DNN overhead. Moderate effort (onnx2tf conversion).

### 6. Async embedding computation
Don't block tracking loop for embeddings — results arrive next frame.

## Device
Xiaomi 13 Pro — Snapdragon 8 Gen 2, Adreno 740 GPU, Hexagon DSP

tuky191:
## Summary

Closes #41. Major tracking speed improvements for stealth/recording mode.

- **GPU delegate for all 9 ML models** — EfficientDet-Lite2 switched from INT8 (CPU) to FP16 (GPU), all other FP32 models on GPU with CPU fallback. Only VitTracker remains on CPU (OpenCV DNN).
- **Adaptive frame skipping** — skip detector every other frame when VitTracker is confident (>0.6, 5+ confirmations). ~50% fewer detector runs while locked.
- **Adaptive analysis interval** — removed fixed 50ms delay from getBitmap loop. Processing time is the natural throttle.
- **MediaPipe 0.10.33** — fixes IMAGE mode timestamp mismatch bug (google/mediapipe#4448) that broke InteractiveSegmenter on rapid calls.
- **Segmenter GPU workaround** — Adreno 740 returns empty masks for crops >100K pixels. Downscale large crops to ~300x300 before GPU segmenter.
- **Loading spinner** — models load on background thread with per-model status ("Loading detector (GPU)...", etc).
- **True stealth mode** — completely blank screen, no UI elements. Tap to exit. Volume-down only control.
- **Subtler haptics** — reduced amplitudes (locked: 20-100, was 60-255; searching: 40, was 80).
- **Dependency bumps** — AGP 8.13.2, CameraX 1.6, TFLite 2.17, Compose BOM 2026.03, MediaPipe 0.10.33, and more.
- **Session limit** — bumped MAX_SESSIONS from 10 to 100.
- **Embedding fallback** — `embedAndCrop(fallback=true)` during search ensures embeddings are always computed even when segmentation fails.

## GPU delegate status

| Model | Delegate |
|---|---|
| EfficientDet-Lite2 (FP16) | GPU (MediaPipe) |
| MobileNetV3 embedder | GPU (MediaPipe) |
| BlazeFace | GPU (MediaPipe) |
| magic_touch segmenter | GPU (MediaPipe) + downscale workaround |
| YOLOv8n, Crossroad-0230, age-gender, OSNet, MobileFaceNet | GPU (TFLite) with CPU fallback |

## Test plan

- [x] `./gradlew testDebugUnitTest` — all tests pass
- [x] Device test: normal mode tracking (mouse, keyboard, person)
- [x] Device test: stealth mode — blank screen, volume-down cycle works
- [x] Device test: recording mode — 4K video capture unaffected
- [x] Device test: outdoor tracking (dog, person) — reacquisition working
- [x] Segmenter masks verified: 100% success rate after downscale fix
- [x] GPU delegate logs confirmed: all 5 TFLite models + 4 MediaPipe models on GPU
- [x] Haptics: noticeably subtler on device


tuky191 (commented):
## PR #42 Review — GPU delegate, FP16 detector, speed optimizations, stealth mode

**1. `NormalizedKeypoint` uses pixel coordinates, not normalized [0,1]**
`ObjectSegmenter.kt:105` — `NormalizedKeypoint.create()` expects values in `[0, 1]` range, but pixel coordinates are being passed (e.g. `segBitmap.width / 2f` = 150.0, not 0.5). This was a pre-existing bug at line 242 too, but the new code at line 105 copies it:

```kotlin
// Bug: pixel coordinates, not normalized
NormalizedKeypoint.create(segBitmap.width / 2f, segBitmap.height / 2f)
```

Should be:
```kotlin
NormalizedKeypoint.create(0.5f, 0.5f)
```

Since you're always centering the keypoint on the crop, `(0.5, 0.5)` is correct regardless of resolution. The fact that this works on device suggests MediaPipe may be interpreting these as pixel coords internally despite the "Normalized" name — but the API contract says [0,1]. Fix both call sites (line 105 and 242).

**2. `GpuDelegate` never closed — resource leak**
`EmbeddingUtils.kt:147-149` — `GpuDelegate` is created but never stored. When the `Interpreter` is closed, the delegate isn't released. Each TFLite GPU model leaks a native GPU context:

```kotlin
val gpuDelegate = GpuDelegate(gpuOptions)
val options = Interpreter.Options().addDelegate(gpuDelegate)
Interpreter(model, options)  // delegate not stored, never closed
```

Fix: either store the delegate alongside each interpreter and close it in `close()`/`shutdown()`, or use a wrapper. 5 TFLite models × 1 leaked delegate each = 5 GPU contexts leaked per app lifecycle.

**3. Race condition: `objectTracker` accessed before initialization**
`CameraViewModel.kt:137` — `lateinit var objectTracker` is initialized on `Dispatchers.Default`, but several methods only check `isTrackerReady` at the start:

- `onTapToLock` (line 140) ✓ guarded
- `onVolumeDown` (line 207) ✓ guarded
- `toggleRecording` — calls `startBitmapAnalysis()` which accesses `objectTracker.processBitmap()` — **NOT guarded** at the point of use
- `clearTracking` (line 265) — calls `objectTracker.clearLock()` — **NOT guarded**
- `toggleStealthMode` — can trigger `toggleRecording()` → `startBitmapAnalysis()`

If the user triggers recording before model loading finishes, `objectTracker` will throw `UninitializedPropertyAccessException`. Add `if (!isTrackerReady) return` to `clearTracking()` and `toggleRecording()`.

**4. `createGpuInterpreter` log doesn't identify which model**
`EmbeddingUtils.kt:150` — All 5 TFLite models log the same message: `"GPU delegate active"`. When debugging on device, you can't tell which model succeeded/failed. Pass the model name:

```kotlin
fun createGpuInterpreter(model: MappedByteBuffer, modelName: String = "unknown", cpuThreads: Int = 2): Interpreter {
    // ...
    Log.i("TFLiteGPU", "GPU delegate active for $modelName")
    // ...
    Log.w("TFLiteGPU", "GPU delegate unavailable for $modelName, using CPU: ${e.message}")
}
```

Then each call site passes its model name (already in companion objects as constants).

**5. Adaptive frame skipping: `vtFrameCounter` not reset on lock clear's visual tracker stop**
`ObjectTracker.kt:165-167` — `clearLock()` resets `vtFrameCounter = 0` ✓, and drift at line 264 resets it ✓, but `prepareForRebind()` at line 152 does NOT reset `vtFrameCounter`. After a stream rebind, the counter carries stale state which could cause an off-by-one in the skip cycle (first frame after rebind might skip detector when it shouldn't).

**6. Bitmap analysis loop: delay moved to else-only branch**
`CameraViewModel.kt:332-334` — The delay is now only when `bitmap == null`. When bitmap is available, the loop runs as fast as `processBitmap()` returns (~30ms). This is intentional per the PR description ("processing time is the natural throttle"), but it means the loop runs unbounded when tracking is fast. On Adreno 740 with GPU delegates, `processBitmap()` could drop below 16ms, causing the loop to compete with the main thread for `getBitmap()`. Consider at minimum a `yield()` to let other coroutines breathe.

**7. `MAX_SESSIONS = 100` — disk usage concern**
`DebugFrameCapture.kt:35` — Bumped from 10 to 100. Each session can contain dozens of PNGs + scenario.json. At 100 sessions × ~5MB/session, that's ~500MB on user storage. Was 10 already aggressive? Consider if 30-50 is a better balance, or add a size-based limit instead.

**8. Stealth mode tap detection: `awaitPointerEvent` loop never restarts**
`CameraScreen.kt:762-772` — After the user taps to exit stealth, the `while(true)` loop breaks. If `toggleStealthMode()` is called again to re-enter stealth, the `pointerInput(Unit)` coroutine has already exited. The `Unit` key means it won't restart. This only matters if stealth is toggled without a recomposition cycle — in practice the `if (!uiState.stealthMode)` branch change triggers recomposition, so this should be fine. But it's fragile.

**9. CLAUDE.md still says `efficientdet-lite2.tflite` (INT8) in the model table**
The model table at CLAUDE.md line ~32 was updated to show INT8 + CPU, but the PR actually replaced it with `efficientdet-lite2-fp16.tflite` on GPU. The table should say FP16 + GPU (MediaPipe), not INT8 + CPU (XNNPACK). The design decision section also says "EfficientDet-Lite2 is the only INT8 quantized model — GPU delegate requires FP32/FP16 input tensors, so it stays on CPU" — but this PR switches it to FP16 GPU. Stale docs.

**10. Dependency versions in CLAUDE.md not updated**
The Dependencies table still shows CameraX 1.4.1, MediaPipe 0.10.21, Compose BOM 2024.12.01, etc. Should be updated to match `libs.versions.toml` (1.6.0, 0.10.33, 2026.03.01).

- GPU fallback pattern is clean and consistent across all 9 models. Try-GPU-catch-CPU with logging. Good.
- Haptic amplitudes (20-100 locked, 40 searching) feel right for non-intrusive feedback. Good call reducing from the old 60-255 range.
- The `onLoadingStatus` callback threading is safe: `_uiState.update` uses `MutableStateFlow.update` which is thread-safe by design.
- `embedAndCrop(fallback=true)` during search is a good improvement — ensures embedding gallery always grows.
- The segmenter downscale workaround (MAX_SEGMENT_PIXELS = 90K) is pragmatic. Comment explains the Adreno 740 issue well.

### Verdict

Fix **#1** (NormalizedKeypoint), **#2** (GpuDelegate leak), and **#3** (race condition guards) before merge. The rest can be addressed in follow-up.

tuky191 (commented):
## PR #42 Follow-up — Fix commit caf7012

| # | Issue | Status |
|---|---|---|
| 1 | NormalizedKeypoint pixel coords → normalized | ✅ Fixed — both call sites now use `(0.5f, 0.5f)`. Comment at line 238 still says "pixel coordinates" — minor, but should say "normalized [0,1] coordinates" to match the other site. |
| 2 | GpuDelegate leak | ✅ Fixed — new `GpuInterpreter` data class pairs interpreter + delegate, `close()` releases both. All 5 TFLite models (YOLOv8, Crossroad-0230, age-gender, OSNet, MobileFaceNet) updated to use `gpu.close()`. Clean pattern. |
| 3 | Race condition on `objectTracker` | ✅ Fixed — `clearTracking()` and `toggleRecording()` now have `if (!isTrackerReady) return` guards. |
| 4 | GPU log doesn't identify model | ✅ Fixed — `modelName` param added, all call sites pass their model name. |
| 5 | `vtFrameCounter` not reset in `prepareForRebind()` | ✅ Fixed — line 153. |
| 6 | Bitmap analysis loop backpressure | ✅ Fixed — `yield()` added after `processBitmap()`. |
| 7 | MAX_SESSIONS=100 | N/A — by request. |
| 9 | CLAUDE.md stale model table | ✅ Fixed — EfficientDet now FP16/GPU, design decision updated, dependency versions updated. |
| 10 | CLAUDE.md stale dependency versions | ✅ Fixed — all versions match `libs.versions.toml`. |

One minor nit: `ObjectSegmenter.kt:240` comment still says "pixel coordinates — Android API" but the code now uses `(0.5f, 0.5f)`. Should say "normalized [0,1] coordinates" to match line 102.

Good to merge.

Issue #43, opened 2026-04-19
Improve tracking responsiveness for fast-moving subjects (kids, pets)


tuky191:
## Problem

When tracking fast-moving subjects (children, pets), the tracker loses focus too frequently. The auto-zoom and re-acquisition are not fast enough to keep up with rapid movement, causing:

1. **VitTracker drift** — loses confidence quickly on fast-moving subjects, triggering unnecessary re-acquisition cycles
2. **Re-acquisition latency** — even with GPU delegates and frame skipping (#41), the search phase takes multiple frames to re-lock
3. **Auto-zoom lag** — zoom adjustments don't keep up with subject movement, resulting in the subject leaving the frame before zoom can pull back

## Use case

Parent filming their child — camera should follow with zoom, keeping the child centered and in focus. The tracking needs to be smooth enough for continuous recording without constant lock/lost/reacquire cycles.

### Tracker resilience
- **Increase VT_MAX_UNCONFIRMED** — currently 10 frames (~0.3s). Fast-moving subjects may not get detector confirmation for longer stretches without actually being lost. Tuning this could reduce false "drift" detections.
- **Lower VT confidence threshold** — currently 0.4. For fast motion, brief low-confidence frames shouldn't kill the tracker.
- **Predictive position** — when VT loses briefly, extrapolate from recent velocity instead of immediately entering search mode.

### Re-acquisition speed
- **Skip segmentation during search** — segmentation is ~10ms per candidate. For 4 candidates that's 40ms. Raw crop embeddings (fallback=true) may be good enough for re-acquisition.
- **Reduce embedding computation** — only compute embedding for same-label candidates, skip wrong-label early.
- **Single-candidate fast path** — if only one same-label candidate exists, skip full scoring and re-lock immediately.

### Auto-zoom responsiveness
- **Increase zoom speed** — currently 0.10 per frame. For fast subjects, this may not keep up.
- **Predictive zoom** — anticipate zoom-out when subject is moving toward frame edge at high velocity.
- **Faster zoom-out on lost** — current `zoomOutForSearch` pulls back 45%. For fast subjects, pull back more aggressively.

### Frame processing
- **Priority processing** — when locked and subject near edge, skip frame skipping and run detector every frame.
- **Reduce processing pipeline** — during stable tracking (VT confident), skip unnecessary steps (label enrichment, person attributes).

## Related

- #41 — GPU delegate + speed optimizations (merged/in review)
- #14 — Manual camera controls (zoom done, recording done)

[...]

Issue #9, opened 2026-04-15
Doc: CLAUDE.md data flow says "hard filter" for label but code uses soft scoring


tuky191:
## Problem

CLAUDE.md describes the re-acquisition data flow in step 3 as:

> `ReacquisitionEngine` scores candidates: position (decays over time) + size + label (hard filter) + appearance similarity (45% weight)

The label hard filter was removed in commit `b0076c1`. Label is now a 20% soft scoring factor, not a hard gate. The parenthetical should say "(soft scoring factor, 20% weight)" or similar.

## Location

`CLAUDE.md`, under **Data Flow** > **Re-acquisition (object lost)**, step 3.

## Fix

Change:
```
label (hard filter)
```
To:
```
label (20% scoring factor)
```

## Context

The label filter evolution is documented correctly in the **Key Design Decisions** section ("Label is a scoring factor, not a gate") and in the **Tunable Parameters** table. Only the data flow summary is stale.

PR #44, opened 2026-04-20
Tracking responsiveness: velocity, SurfaceTexture recording, async embeddings (#43)

tuky191:
## Summary

Major tracking performance overhaul for fast-moving subjects during recording:

- **SurfaceTexture frame reader** — replaces `PreviewView.getBitmap()` during recording. Custom GL pipeline reads PRIV stream at 29fps, processes at 15fps. Maintains 4K video (PRIV+PRIV) while providing analysis frames 2× faster than before (15fps vs 7-10fps)
- **Decoupled viewfinder** — viewfinder updates at GL rate (~29fps) independent of processing thread. No more choppy preview during search mode
- **Async embedding pipeline** — embeddings compute on background thread while detector processes next frame. Results merged by IoU across frames. 3/5 reacquisitions now use embedding identity (sim 0.42-0.87)
- **Velocity-aware tracking** — EMA-smoothed velocity estimation drives adaptive drift detection (3/5/10 frame thresholds), predicted search position, and velocity-aware FTF matching
- **Embedding floor gate** — `MIN_EMBEDDING_SIMILARITY=0.15` prevents accepting candidates the embedder says are different objects (fixes person swap bug)
- **Embedding-required gate** — blocks reacquisition without embeddings for first 10 search frames when gallery exists
- **Gradual zoom-out** — 5-frame delay + 15%/frame pullback replaces jarring immediate 45% pullback
- **Video replay integration tests** — feed real video through full ObjectTracker on device with natural frame dropping

## Test plan

- [x] `./gradlew testDebugUnitTest` — 184 tests pass (includes new VelocityEstimator, ZoomController, ScenarioReplay tests)
- [x] Device test: recording produces 4K video (3840×2160) with smooth viewfinder
- [x] Device test: GL thread 29fps, processing 15fps confirmed in logs
- [x] Device test: reacquisition uses embeddings (override at sim=0.837)
- [x] Device test: wrong-category candidates (tv, keyboard, laptop) correctly rejected
- [x] Person swap scenario: boy/wife discrimination via embedding floor gate
- [ ] Outdoor test with fast-moving subject (toddler)


tuky191 (commented):
## PR #44 Review — Tracking responsiveness: velocity, SurfaceTexture recording, async embeddings

Large PR (1685 additions). Well-structured: clear separation between GL pipeline, velocity estimation, async embeddings, and reacquisition engine changes. Tests are thorough (VelocityEstimator 14 tests, ZoomController 4 tests, ScenarioReplay 7 tests, plus instrumented VideoReplayTest).

**1. Async embedding IoU merge can silently assign wrong embeddings**
`ObjectTracker.kt:384-402` — The merge picks the `maxByOrNull` cached result for each detection, but doesn't remove matched entries. If 3 current detections match the same cached entry (e.g. a person standing still, 3 overlapping boxes), all 3 get that embedding. Worse: a high-IoU match with one detection can be the "best" for another detection too. This needs a greedy assignment (mark used entries), same pattern as `FrameToFrameTracker.assignIds()`:

```kotlin
val usedCacheKeys = mutableSetOf<Int>()
val merged = withLabels.map { obj ->
    val bestMatch = cachedEntries
        .filter { (id, result) -> result.cachedBox != null && id !in usedCacheKeys }
        .maxByOrNull { (_, result) -> computeIou(obj.boundingBox, result.cachedBox!!) }
    val iou = bestMatch?.let { computeIou(obj.boundingBox, it.value.cachedBox!!) } ?: 0f
    if (bestMatch != null && iou > 0.3f) {
        usedCacheKeys.add(bestMatch.key)
        // ... copy embeddings
    } else obj
}
```

**2. `SurfaceTextureFrameReader` doesn't release EGL resources on early failure**
`SurfaceTextureFrameReader.kt:802-804` — If the GL thread throws between `initEGL()` and the main render loop (e.g. texture creation fails), the catch block logs the error and unblocks the latch but never calls `releaseEGL()`. The EGL display/context/surface leak. Add `releaseEGL()` to the catch block.

**3. Viewfinder bitmap memory pressure — no recycling**
`CameraViewModel.kt:173` — Comment says "Don't recycle — Compose may still be drawing previous frames. Let GC handle." At 29fps of 480×640 ARGB_8888 bitmaps = ~1.2MB each × 29/s = ~35MB/s of GC pressure. This will cause GC pauses on lower-end devices. Consider:
- Use a ring buffer of 2-3 pre-allocated bitmaps with `copyPixelsFromBuffer` (reuse instead of allocate)
- Or at minimum, recycle the *previous* value from `_viewfinderBitmap` when setting the new one (the old one is guaranteed to not be drawn anymore since StateFlow only exposes latest)

Actually, looking closer: the GL thread creates the bitmap, copies it for viewfinder (`bitmap.copy`), then stores the original in `latestFrame`. The viewfinder copy is sent to `_viewfinderBitmap.value` which replaces the previous. The previous value IS safe to recycle since Compose only draws the current StateFlow value. But you'd need to capture the old value before updating:
```kotlin
cameraManager.onViewfinderFrame = { bitmap ->
    val old = _viewfinderBitmap.getAndSet(bitmap) // needs AtomicReference
    old?.recycle()
}
```
This is medium-impact if the device has 8GB+ RAM, but high-impact on 4GB devices.

**4. `GEOMETRIC_OVERRIDE_THRESHOLD = 0.55` is dangerously close to noise floor**
`ReacquisitionEngine.kt:351` — The new tiered thresholds split geometric override (0.55) from label override (0.70). But for MobileNetV3 1280-dim embeddings on visually similar objects (two cups, two people with same clothes), cosine similarity can sit in the 0.4-0.6 range. A 0.55 threshold means a different-but-similar object can bypass position/size gates. The old single threshold of 0.7 was safer. Consider 0.60-0.65 as a compromise.

**5. `EMBEDDING_REQUIRED_FRAMES = 10` blocks reacquisition too long when embeddings fail**
`ReacquisitionEngine.kt:358-359` — If the async pipeline consistently fails to produce embeddings (e.g. segmenter GPU issue), candidates are rejected for 10 frames (~333ms) every time, creating a dead zone. The comment says "embeddings typically arrive within 2-3 frames" — so 5 frames would be a safer limit with the same protection. Or track consecutive embedding failures and disable the gate after N failures.

**6. Velocity prediction drifts search area off-screen for long-lost objects**
`ReacquisitionEngine.kt:341-346` — `predictedCenterX = refBox.centerX() + velocityX * framesLost`. At framesLost=100 with velocity=0.03, that's 3.0 → clamped to 1.0. The prediction pins to the edge of the frame, which is wrong — the position should revert to the original reference position as uncertainty grows. Consider capping the prediction to e.g. `min(framesLost, 10) * velocity` (same timeframe as the embedding required gate).

**7. `embeddingExecutor` thread not named**
`ObjectTracker.kt:60` — `Executors.newSingleThreadExecutor()` creates an unnamed thread. For debugging with thread dumps, use a named factory:
```kotlin
private val embeddingExecutor = Executors.newSingleThreadExecutor { r -> Thread(r, "EmbeddingAsync") }
```

**8. `needEmbeddings` drops the `framesLost==0` case**
`ObjectTracker.kt:349-350` — Before this PR:
```kotlin
val needEmbeddings = reacquisition.hasEmbeddings && 
    (reacquisition.isSearching || (reacquisition.isLocked && reacquisition.framesLost == 0))
```
After:
```kotlin
val needEmbeddings = reacquisition.hasEmbeddings && 
    reacquisition.isSearching && !hasDirectMatch
```
The `framesLost==0` case (just locked, need to verify via embedding on next frame) is gone. This means after a visual tracker handoff to detector path, if the locked ID happens to be present, embeddings are never computed for verification. Was this intentional? If the direct-ID match is sufficient, fine — but it weakens the cross-check that catches ID-assignment errors from `FrameToFrameTracker`.

**9. `VideoReplayTest` relies on device-local files without CI integration**
The instrumented tests require `adb push test_videos/` to the device. The JSON specs are in the repo but the MP4s aren't (`.gitignore` has `*.mp4`). Document the expected test data flow and where the canonical videos live, or they'll rot.

**10. `assignIdsWithVelocity` duplicates most of `assignIds` logic**
`FrameToFrameTracker.kt:17-69` — The velocity-aware version is 53 lines that duplicate `assignIds` except for (a) shifting boxes and (b) using `minIouWithPrediction`. Could be unified with an optional velocity parameter to reduce maintenance surface. Not urgent but both methods need to stay in sync.

**11. Scenario JSON files committed as single-line blobs**
`app/src/test/resources/scenarios/boy_indoor_wife_swap.json` etc. are likely multi-KB single-line files. These are fine for test resources but impossible to diff in reviews. Consider adding a note that they're machine-generated and shouldn't be hand-edited.

**12. `CameraScreen.kt:57` — `import androidx.compose.foundation.Image`**
This should be `androidx.compose.foundation.Image` function, which is correct in Compose. Just confirming — the import compiles since tests pass.

- The SurfaceTexture → GL → FBO → glReadPixels pipeline is the standard approach for camera frame interception. Clean implementation with proper Y-flip via quad vertices instead of a second pass.
- Velocity EMA with α=0.3 is reasonable — responsive enough to detect direction changes in 3-4 frames, smooth enough to filter jitter.
- The embedding floor gate (MIN_EMBEDDING_SIMILARITY=0.15) is a good addition — prevents the "person swap" failure mode where re-ID or attributes rescue a clearly-different object.
- Gradual zoom-out (5-frame delay + 15%/frame) is noticeably better UX than the jarring 45% pullback.
- The 3 existing tests updated to account for MIN_EMBEDDING_SIMILARITY floor (orthogonal embeddings now above 0.15) — good, keeps test invariants honest.

### Verdict

Fix **#1** (greedy IoU merge) and **#2** (EGL leak on early failure). **#3** (viewfinder GC pressure) is borderline high — fine on Xiaomi 13 Pro but could be bad on 4GB devices. Address **#4-#8** as judgment calls. The rest can be follow-up.

tuky191:
**Correction to review:** Disregard **#9** (VideoReplayTest MP4s not in repo). That's by design — videos stay off GitHub. The remaining 11 items stand.

tuky191 (commented):
## PR #44 Follow-up — Fix commit b723711

| # | Issue | Status |
|---|---|---|
| 1 | Async embedding IoU merge not greedy | ✅ Fixed — `usedCacheKeys` set prevents duplicate assignments. Same greedy pattern as FTFTracker. |
| 2 | EGL leak on early GL thread failure | ✅ Fixed — `releaseEGL()` called in catch block with `eglDisplay != EGL_NO_DISPLAY` guard. |
| 3 | Viewfinder bitmap GC pressure | ✅ Fixed — previous bitmap recycled before setting new StateFlow value. ~35MB/s → ~1.2MB/s. |
| 4 | `GEOMETRIC_OVERRIDE_THRESHOLD` too low (0.55) | ✅ Raised to 0.60 — safer distance from similar-object noise floor. |
| 5 | `EMBEDDING_REQUIRED_FRAMES` too long (10) | ✅ Reduced to 5 — matches the "2-3 frames typical" with margin. |
| 6 | Velocity prediction drifts off-screen | ✅ Fixed — `predictionFrames = minOf(framesLost, 10)` caps prediction range. |
| 7 | `embeddingExecutor` unnamed thread | ✅ Named `"EmbeddingAsync"`. |
| 8 | `needEmbeddings` dropped `framesLost==0` case | Not addressed — assuming intentional since direct ID match is sufficient for the detector path. |
| 9 | VideoReplayTest MP4s not in repo | N/A — by design. |
| 10 | `assignIdsWithVelocity` duplication | Not addressed — low priority. |
| 11 | Scenario JSON single-line blobs | Not addressed — low priority. |

One concern with #3 fix: `_viewfinderBitmap.value` is a `MutableStateFlow`. Setting `.value` on GL thread while Compose reads on main thread — `StateFlow` is thread-safe for value updates, but the `old?.recycle()` call happens on the GL callback thread. If Compose is mid-draw using `old` when `recycle()` is called, you'll get a `RuntimeException: Canvas: trying to use a recycled bitmap`. The window is tiny (StateFlow collectors see new value almost instantly), but it's a race. Safer: post the recycle to the next GL frame via a `toRecycle` field, or use `_viewfinderBitmap.getAndUpdate { bitmap }` to atomically swap and recycle the returned old value. Worth a quick think — if you haven't seen crashes on device, it's likely fine in practice.

Good to merge (with the caveat above as a known risk).

Issue #45, opened 2026-04-20
Tracking responsiveness: remaining improvements after PR #44


tuky191:
## Context

PR #44 addressed the core tracking responsiveness issues (SurfaceTexture recording, velocity estimation, async embeddings, embedding gates). This issue now focuses on the **VT drift detection** subproblem (point 8 from the original list), which was elevated to primary focus after analysis during #53.

## Problem: Drift detection conflates two different failure modes

**Current drift detection** (`ObjectTracker.kt` ~line 247-307):
1. Every frame, run both VT and detector
2. Check if any detection's bounding box overlaps VT's box (same category IoU > 0.15, any category IoU > 0.4) — if yes, "confirmed"
3. If not confirmed for N consecutive frames (10, or adaptive 3-10 based on velocity), treat as drift and kill VT

**The bug:** Step 2 conflates two fundamentally different scenarios:
- **True drift**: VT slid onto a different object. Detector finds something at VT's position that doesn't match. → Kill VT.
- **Detector absence**: Detector missed our object (common for small/uniform objects, motion blur, label flicker). Object is still there, VT is correct. → Should NOT kill VT.

**Real example (bowl session 2026-04-23)**: Tracking a yellow bowl in a bathroom. A potted plant sits 10cm away. Detector sometimes finds only "potted plant" (not "bowl"). The potted plant's box is above the bowl — no overlap. Current logic: "detector found something but nothing overlaps VT" → increment unconfirmed → DRIFT at frame 10. But the bowl is right there, VT is tracking correctly.

**Speed logs confirm**: `DRIFT detected — 11 unconfirmed frames (max=10, speed=0.0009)` — speed ≈ 0 means VT isn't drifting at all, it's stationary on the target.

## Research findings (2026-04-23)

Every mature MOT/SOT tracker (DeepSORT, ByteTrack, OC-SORT, TLD, DaSiamRPN, LTMU, Stark, Norfair) treats **the tracker's own confidence as the primary drift signal**, with the detector playing a supporting/peer role. Our current design has it inverted: the detector (higher false-negative rate on uniform objects) vetoes the tracker (lower false-negative rate when locked).

### Key papers and techniques

- **TLD (Kalal et al., PAMI 2012)** — classical formulation of this exact problem. Their P-expert explicitly identifies missed detections as detector false negatives, not drift. "The detector is a peer of the tracker, not a judge of it."
[...]

Issue #7, opened 2026-04-15
Minor: DebugFrameCapture allocates Paint objects on every capture


tuky191:
## Problem

`DebugFrameCapture.drawEventBanner()` creates two new `Paint` objects per call. Other paints in the class (`lockedPaint`, `candidatePaint`, `reacquiredPaint`, `lostPaint`, `labelPaint`, `labelBgPaint`) are pre-allocated as fields.

## Location

`app/src/main/java/com/haptictrack/tracking/DebugFrameCapture.kt` lines 167-177.

## Current code

```kotlin
private fun drawEventBanner(canvas: Canvas, text: String, frameWidth: Float) {
    val bannerPaint = Paint().apply {    // allocated every call
        color = Color.BLACK
        alpha = 200
    }
    val textPaint = Paint().apply {      // allocated every call
        color = Color.YELLOW
        textSize = 36f
        isAntiAlias = true
        typeface = Typeface.DEFAULT_BOLD
    }
    // ...
}
```

## Fix

Move these to class-level fields alongside the other paints:

```kotlin
private val bannerBgPaint = Paint().apply {
    color = Color.BLACK
    alpha = 200
}

private val bannerTextPaint = Paint().apply {
    color = Color.YELLOW
    textSize = 36f
    isAntiAlias = true
    typeface = Typeface.DEFAULT_BOLD
}
```

Then use them in `drawEventBanner()`.

## Impact

Negligible performance impact since debug captures are infrequent (a few per tracking event). This is a consistency fix — the class already pre-allocates six other Paint objects, so these two should follow the same pattern.

tuky191:
Fixed in PR #10

Issue #6, opened 2026-04-15
Design: Duplicate cosineSimilarity and bestGallerySimilarity implementations


tuky191:
## Problem

Cosine similarity is implemented twice — once in `ReacquisitionEngine` and once in `AppearanceEmbedder`. Similarly, `bestGallerySimilarity()` in `ReacquisitionEngine` duplicates `bestSimilarity()` in `AppearanceEmbedder`.

## Locations

**ReacquisitionEngine.kt** lines 297-307:
```kotlin
internal fun bestGallerySimilarity(candidateEmbedding: FloatArray): Float {
    if (embeddingGallery.isEmpty()) return 0f
    return embeddingGallery.maxOf { cosineSimilarity(it, candidateEmbedding) }
}

private fun cosineSimilarity(a: FloatArray, b: FloatArray): Float {
    if (a.size != b.size) return 0f
    var dot = 0f
    for (i in a.indices) dot += a[i] * b[i]
    return dot
}
```

**AppearanceEmbedder.kt** lines 123-139:
```kotlin
fun bestSimilarity(candidate: FloatArray, gallery: List<FloatArray>): Float {
    if (gallery.isEmpty()) return 0f
    return gallery.maxOf { cosineSimilarity(candidate, it) }
}

fun cosineSimilarity(a: FloatArray, b: FloatArray): Float {
    if (a.size != b.size) return 0f
    var dot = 0f
    for (i in a.indices) { dot += a[i] * b[i] }
    return dot
}
```

## Why it matters

- If the similarity algorithm ever needs to change (e.g., adding epsilon for numerical stability, or switching from dot product to full cosine), it must be changed in two places.
- `ReacquisitionEngine` is a pure logic class with no Android dependencies — it should not need to know how similarity is computed.

## Fix

Keep `cosineSimilarity` in one place. Options:

**Option A** — Top-level function in a shared file (e.g., `EmbeddingUtils.kt`):
```kotlin
// tracking/EmbeddingUtils.kt
fun cosineSimilarity(a: FloatArray, b: FloatArray): Float { ... }
fun bestSimilarity(candidate: FloatArray, gallery: List<FloatArray>): Float { ... }
```

**Option B** — Keep it on `AppearanceEmbedder` (it owns the embedding domain) and have `ReacquisitionEngine` accept a similarity function via constructor:
```kotlin
class ReacquisitionEngine(
    // ...
[...]

PR #47, opened 2026-04-21
Stronger object identity: resolution, centroid, negative rejection (#45)

tuky191:
## Summary

Major identity and tracking stability overhaul. Replaces per-label matching with embedding-first identity, fixes VT stability, and adds multiple regression tests from real device sessions.

### Identity
- **Analysis resolution 480x640 -> 640x854** - 78% more pixels. Embedding similarity jumps from 0.5-0.6 to 0.83-0.88. Processing stays at 14fps.
- **Person/not-person category gate** - replaces per-label matching. Labels are unreliable. Binary gate: person only accepts person, everything else accepts any non-person. Embedding handles identity.
- **Embedding centroid** - L2-normalized mean of gallery. More stable than max-of-gallery.
- **Adaptive embedding floor** - mature gallery: floor=0.45, immature: floor=0.15. Prevents keyboard (sim=0.318) accepted when tracking mouse.
- **Mature gallery never accepts without embeddings** - no timeout fallback with >=8 entries.
- **Negative rejection** - penalize candidates closer to negatives than to centroid.
- **Relative scoring** - clear embedding margin over second-best accepts at lower threshold.
- **Faster gallery diversification** - every 0.5s with diversity check.

### VT stability
- **VT confirmation by position, not label** - label flicker was killing VT every 1-3s. Now holds 20+ seconds.
- **Size-adaptive confidence floor** - small ROIs get 0.25 floor instead of 0.4.
- **Detector contradiction vs non-confirmation** - empty detector output no longer counts against VT.

### Reacquisition
- **Single-candidate fast path** - bypass embedding wait, but only with immature gallery.
- **Stricter size gate without embeddings** - cap at 3x when gallery is mature.
- **YOLOv8 dropped from search** - saves ~270ms per call. Only at lock time.

### UX
- **Slower zoom-out** - 8-frame delay + 8%/frame (was 5 + 15%).
- **Auto-start recording on lock** - tap or volume-down both lock + record immediately.

## On-device video replay results

| Test | Before | After | Threshold |
|------|--------|-------|-----------|
| boy_indoor_wife_swap | 84% | 93% | >=30% |
| chair_living_room | new | 83% | >=70% |
| flowerpot_wrong_reacq | new | 90% | >=65% |

## Test plan

- [x] 210 unit tests pass
- [x] On-device: boy 93%, chair 83%, flowerpot 90%
- [x] VT holds 20+ seconds on stationary bowl (was 1-3s)
- [x] Keyboard rejected at sim=0.318 (floor 0.45)
- [x] 14fps processing, 29fps GL at 640x854


tuky191 (commented):
## PR #47 Review — Stronger object identity: resolution, centroid, negative rejection

438 additions, 193 deletions. This is a significant identity architecture shift: per-label matching → binary person/not-person gate + embedding-first identity. Well-motivated by real failures (label flicker killing VT, keyboard accepted when tracking mouse).

**1. `addNegativeExample` stores embedding of gate-rejected candidate — but only for cross-category**
`ReacquisitionEngine.kt:453` — Negative examples are only added when the person/not-person gate rejects a candidate with an embedding. But the main failure mode (keyboard vs mouse, both non-person) passes the gate and is only caught by the embedding floor. Candidates rejected by the embedding floor (line 411) should also be added as negatives — those are the ones the system actually confused:

```kotlin
if (hasAppearance && appearanceScore < embeddingFloor) {
    addNegativeExample(candidate.embedding!!)
    return null
}
```

Without this, `negativePenalty()` only sharpens the person/non-person boundary (which the gate already handles) and never helps within the same category.

**2. Relative scoring can accept below `minScoreThreshold`**
`ReacquisitionEngine.kt:356` — The relative margin path accepts at `minScoreThreshold * 0.8f`:
```kotlin
if (margin > 0.2f && bestSim > 0.4f && best.second >= minScoreThreshold * 0.8f) {
    return best.first  // bypasses minScoreThreshold
}
```
At `minScoreThreshold=0.45`, this accepts at 0.36. A candidate scoring 0.36 with only 0.2 margin over the next could be two bad candidates where one is slightly less bad. The `bestSim > 0.4f` guard helps, but consider tightening to `minScoreThreshold * 0.9f` to keep the safety margin. This is a judgment call — flag it.

**3. `+ 0f` dead code in all 4 scoring tiers**
`ReacquisitionEngine.kt:520, 538, 558, 570` — `labelBonus` was replaced with `+ 0f` literal. This is dead arithmetic. Either remove the `+ 0f` entirely (cleaner) or leave a `// no label bonus` comment. Currently it reads like a placeholder for something that should be there.

**4. VT confirmation drops label check — widened blast radius**
`ObjectTracker.kt:253-258` — VT is now confirmed by any detection overlapping its position, regardless of label. This fixes the label flicker problem (bowl→potted plant killing VT), but it also means VT is confirmed by a completely different object that happens to occupy the same box (e.g. person walks in front of the tracked cup, large person box overlaps cup box with IoU > 0.15). The IoU threshold of 0.15 is very low. Consider requiring either:
- Same label, OR
- IoU > 0.4 (strong overlap = same object regardless of label)

This preserves the label-flicker fix while preventing confirmation by a different-category overlapping object.

**5. `detectorFoundSomething` logic has a subtle gap**
`ObjectTracker.kt:260-261` — `val detectorFoundSomething = skipDetector || tracked.isNotEmpty()`. When `skipDetector` is true, `detectorFoundSomething` is true, but the `confirmed` var is also forced to true (line 253). So `detectorFoundSomething` only matters when `!skipDetector`. In that case, if `tracked.isEmpty()`, unconfirmed doesn't increment — good. But what about the edge case where the detector finds a single large background detection (e.g. "dining table" covering the whole frame) that overlaps everything at IoU > 0.15? That confirms VT on a drifted position. This is the same concern as #4.

**6. Auto-start recording on tap — accidental recordings likely**
`CameraViewModel.kt:173-174` — Every `onTapToLock` now calls `toggleRecording()`. This means any exploratory tap-to-track (just checking if detection works, not wanting to record) starts a recording immediately. Consider making this opt-in or at least adding a short delay / confirmation state. Volume-down auto-record makes sense (deliberate gesture), but tap is often exploratory.

**7. Adaptive embedding floor (0.45 for mature gallery) — no test coverage**
The floor jump from 0.15 to 0.45 at gallery size ≥8 is a critical threshold that determines whether candidates pass or fail. There's no unit test verifying this behavior. Add a test that: (1) locks with <8 embeddings, verifies floor is 0.15; (2) adds embeddings to reach 8, verifies floor is 0.45; (3) confirms a candidate at 0.3 similarity passes with immature gallery but fails with mature.

**8. `centroidSimilarity` returns 0 when gallery is empty — affects `embScore` blend**
`ReacquisitionEngine.kt:257-258` — When gallery has 1 entry, centroid = that entry (same as best-of-gallery). When gallery is empty, centroid is null, returns 0f. This makes `embScore = (appearanceScore + 0) / 2`, halving the embedding score. This shouldn't happen in practice (hasAppearance implies gallery is non-empty), but it's fragile.

**9. `negPenalty` not applied in face or re-ID tiers**
`ReacquisitionEngine.kt:514-520, 532-538` — Negative penalty only applies in the "generic embedding" tier (line 552-558). Face and re-ID tiers add `+ 0f` where `labelBonus` was. If the same person appears as a negative (e.g. wife rejected, then reappears), the face/re-ID tier won't penalize her. Probably fine since face/re-ID are strong signals, but worth noting.

**10. `lockedIsPerson` set from COCO label only**
`ReacquisitionEngine.kt:135` — `lockedIsPerson = (cocoLabel ?: label) == "person"`. If YOLOv8 enriches "person" to "Man"/"Woman"/"Boy"/"Girl" at lock time, and `cocoLabel` is passed correctly, this works. But if only the enriched label is passed (without cocoLabel), a locked "Boy" would be `lockedIsPerson = false`. Verify that `lockOnObject` always passes the COCO label as `cocoLabel`.

**11. ScenarioReplayTest cup reacquisition assertion weakened**
`ScenarioReplayTest.kt:140` — Changed from `assertEquals("cup", event.label)` to `assertNotEquals("person", event.label)`. This is correct for the new person/not-person gate, but now allows reacquisition as "laptop", "bed", etc. which was previously prevented. The trade-off is intentional (embedding handles identity), but worth noting that scenario replay tests are now looser.

**12. ZoomController test constants updated but test names not**
`ZoomControllerTest.kt` — Delay changed from 5→8 frames and pullback from 15%→8%. The test names should reflect the new values if they mention specific numbers.

- The binary person/not-person gate is a pragmatic simplification. Label flicker between non-person objects (bowl/potted plant, deer/sheep) was causing real tracking failures. Letting embedding handle within-category identity is sound.
- Centroid as `L2-normalized mean` is the right approach — more stable than max-of-gallery for cosine similarity comparisons. The `computeCentroid` implementation is clean.
- VT stability improvements (position-only confirmation, size-adaptive confidence floor) should dramatically improve small-object tracking. The 0.25 floor for small ROIs is a good heuristic.
- Gallery diversification at 0.5s with centroid-sim < 0.92 diversity check is smart — prevents the gallery from filling with near-identical embeddings.
- Resolution bump 480→640 for 78% more pixels with same fps is a free win.
- 210 tests (up from 184) with good coverage of the new person/not-person gate semantics.

### Verdict

Fix **#1** (add negatives from embedding floor rejects) — this is the main gap in the negative example system. **#2** and **#3** are easy cleanups. **#4** (VT confirmation without label) is the riskiest behavior change — monitor on device. **#6** (auto-record on tap) is a UX concern worth discussing. **#7** (missing test for adaptive floor) should be added. The rest can be follow-up.

tuky191 (commented):
## PR #47 Follow-up — Fix commit e2b6cf7

| # | Issue | Status |
|---|---|---|
| 1 | Negatives only from cross-category, not embedding floor | ✅ Fixed — `addNegativeExample()` now called at embedding floor reject (line 411). |
| 2 | Relative scoring accepts at 0.8x threshold | ✅ Tightened to 0.9x. |
| 3 | `+ 0f` dead code in all tiers | ✅ Removed — face/re-ID tiers clean, generic tier now `- negPenalty` directly. |
| 4 | VT confirmation too permissive (any overlap at 0.15) | ✅ Fixed — same-category at IoU > 0.15, cross-category only at IoU > 0.4. Clean solution. |
| 5 | `detectorFoundSomething` edge case | Addressed by #4 — high-IoU requirement for cross-category prevents false confirmation by overlapping background objects. |
| 6 | Auto-record on tap | ✅ Removed — tap locks without recording, volume-down locks+records. Good call. |
| 7 | No test for adaptive embedding floor | ✅ Added — 2 tests: immature gallery passes sim=0.3, mature gallery rejects low-sim. |
| 8 | `centroidSimilarity` returns 0 when empty | Not addressed — acknowledged as theoretical only (hasAppearance implies non-empty gallery). |

All clean. Good to merge.

PR #54, opened 2026-04-23
Unify on SurfaceTexture pipeline, remove 3-stream mode (#48)

tuky191:
## Summary

- Always use 2-stream SurfaceTexture pipeline (Preview + VideoCapture). Removed the dual-mode architecture where tracking used 3-stream (Preview + ImageAnalysis + VideoCapture) and recording switched to 2-stream.
- Recording now just toggles VideoCapture on/off — no `unbindAll()`, no `prepareForRebind()`, no camera rebind. Eliminates #46 entirely.
- Stealth mode is purely a UI overlay change — no camera reconfiguration needed.
- Removed ~163 lines of dead code across 4 files: `ImageAnalysis` setup, `enterRecordingMode()`/`exitRecordingMode()`, `processImage(ImageProxy)`, `startBitmapAnalysis()`/`stopBitmapAnalysis()`, mode flags, and conditional viewfinder rendering.
- Updated CLAUDE.md to reflect the new architecture.

## Files changed

| File | Changes |
|------|---------|
| `CameraManager.kt` | Removed dual-mode logic, `imageAnalysis`, `recordingMode`/`analysisActive` flags. Always creates SurfaceTextureFrameReader. Renamed `onRecordingFrame` → `onAnalysisFrame`. |
| `CameraViewModel.kt` | Removed mode switching, `prepareForRebind()` for recording, dead code (`startBitmapAnalysis`, `processImage`). `toggleRecording()` just starts/stops. `toggleStealthMode()` is purely UI. |
| `CameraScreen.kt` | Viewfinder always renders from `viewfinderBitmap` (removed recording-only conditional). |
| `ObjectTracker.kt` | Removed `analyzer` property, `processImage(ImageProxy)`, `imageProxyToBitmap()`, ImageAnalysis imports. |
| `CLAUDE.md` | Replaced 3-stream/ImageAnalysis/mode-switching docs with unified pipeline description. |

## Test plan

- [x] 225+ unit tests pass
- [x] On-device video replay: 10/12 pass, 2 pre-existing flaky failures (mouse rotation detector coverage #53, person playground GPU non-determinism — passes on re-run)
- [x] Manual device test: tracking without recording, start/stop recording (seamless), stealth mode, volume-down cycle, camera switch — all working
- [x] SurfaceTexture confirmed active: `STFrameReader: GL thread produced 2069 frames`

Closes #48, closes #46

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #52, opened 2026-04-22
Object identity: learn what we're tracking (#50)

tuky191:
## Summary

- **Segmenter rewrite**: raw TFLite GPU replaces broken MediaPipe InteractiveSegmenter (13-17ms vs 110ms)
- **Online classifier**: lightweight logistic regression trained on-device from gallery + scene negatives, learns per-object decision boundary
- **Structural anti-fluke defenses**: tentative confirmation (DeepSORT), ratio test (SIFT), classifier-gated bypass, gallery-relative threshold
- **Search zoom cap**: prevents zooming to 1x during search which makes small objects undetectable

Zero wrong reacquisitions across all test scenarios (chair, mouse, person, boy, flowerpot). Keyboard/couch/bed/laptop all correctly rejected.

## Key changes

| File | Changes |
|------|---------|
| `ObjectSegmenter.kt` | Complete rewrite: MediaPipe → raw TFLite GPU |
| `ReacquisitionEngine.kt` | Prototype scoring, strict embedding gate, tentative confirmation, ratio test, classifier-gated bypass, gallery-relative threshold |
| `EmbeddingUtils.kt` | Online logistic classifier, `minPairwiseSimilarity()` |
| `ObjectTracker.kt` | Scene negative collection during VT tracking |
| `ZoomController.kt` | Search zoom-out capped at 50% of locked zoom |
| `VideoReplayTest.kt` | New mouse_desk_rotation scenario (phone rotation identity test) |
| `ReacquisitionEngineTest.kt` | 10+ new tests: tentative, ratio, classifier, gallery-relative |
| `ScenarioReplayTest.kt` | Threshold adjustments for tentative confirmation latency |

## Test plan

- [x] 220+ unit tests pass (`./gradlew testDebugUnitTest`)
- [x] On-device video replay: chair_living_room — 0 wrong reacqs (was 3-4)
- [x] On-device video replay: mouse_desk_rotation — 0 wrong reacqs during phone rotation
- [x] On-device video replay: all 5 existing scenarios pass (flowerpot flaky due to pre-existing bitmap bug)
- [x] Manual testing: mouse tracking with phone rotation, all reacquisitions correct
- [ ] Review classifier training performance on longer sessions

Closes #50
Remaining improvements spun off to #51

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #55, opened 2026-04-24
Fix VT drift detection + embedding delivery (#45, #53)

### LINKED ISSUE #45: Tracking responsiveness: remaining improvements after PR #44

### LINKED ISSUE #53: Full-screen phantom detections and zoom-out after phone rotation
## Problem

After phone rotation, EfficientDet produces near-full-screen phantom detections (airplane, book, frisbee) that cover 50-70% of the frame. Combined with search zoom-out, this creates a visual effect where "the selection selects the whole screen."

More critically, **correct detections are hard-rejected for tens of seconds** because the async embedding pipeline can't deliver embeddings during rotation.

Reproduced consistently during bowl tracking in a bathroom — rotating the phone triggers VT loss, search zoom-out, and long reacquisition gaps even when the target is clearly visible.

### Primary: Embedding delivery failure during rotation

The async embedding pipeline computes embeddings one frame behind, then merges them into the current frame by **IoU > 0.3** with the previous frame's boxes. During rotation this breaks completely:

1. Phone rotation causes large frame-to-frame positional shifts in normalized coordinates
2. **FTFTracker loses ID continuity** — target gets new IDs each frame (233→234→241→242→243) because IoU < 0.2 between rotated frames
3. **Embedding cache can't merge** — previous frame's embedding was for a box at a completely different position, IoU < 0.3 with current frame
4. `candidate.embedding == null` → **hard rejected** by the embedding gate (`ReacquisitionEngine.kt:513`)
5. Repeat until phone stops rotating and the target stabilizes

The embedding gate was added in PR #52 to prevent wrong-category reacquisitions:

```kotlin
// ReacquisitionEngine.kt:513
if (hasEmbeddings && candidate.embedding == null) {
    return null  // hard reject — no embedding, can't verify identity
}
```

This gate is correct in principle — accepting without identity verification causes wrong-object locks. But it becomes a trap during rotation because the async pipeline structurally cannot deliver embeddings when positions shift rapidly.

**Evidence from bowl session:** The bowl is detected as "bowl" multiple times during a 73-frame (9 second) search gap, but every detection is rejected with `sim=n/a (hard threshold)` because no embedding was available:

```
162503_214 [Reacq]   rejected id=233 label="bowl" sim=n/a (hard threshold)
162503_756 [Reacq]   rejected id=233 label="bowl" sim=n/a (hard threshold)
162504_693 [Reacq]   rejected id=234 label="bowl" sim=n/a (hard threshold)
...
162509_282 [Reacq]   rejected id=241 label="bowl" sim=n/a (hard threshold)
162509_993 [Reacq]   rejected id=242 label="bowl" sim=n/a (hard threshold)
162510_497 [Reacq]   rejected id=243 label="bowl" sim=n/a (hard threshold)
```

### Secondary: Full-screen phantom detections

EfficientDet produces near-full-screen detections during rotation/motion blur:

| Time | Label | Box | Area |
|------|-------|-----|------|
| 16:25:48 | airplane | [0.00, 0.00, 1.00, 0.70] | 0.70 |
| 16:25:49 | airplane | [0.00, 0.00, 0.99, 0.68] | 0.67 |
| 16:25:02 | book | [0.12, 0.22, 1.00, 0.99] | 0.68 |

`DetectionFilter.maxBoxArea = 0.7` lets these through. Also, `DetectionFilter` runs AFTER `ReacquisitionEngine.processFrame()` (ObjectTracker.kt:501 vs :469), so the engine receives unfiltered detections.

### Tertiary: VT dies too frequently during rotation

9 LOST/REACQUIRE cycles in 1m37s. Most are legitimate (bowl leaves frame during rotation), but some occur while the bowl is clearly visible — likely VT confidence drop during brief rotation transients. Overlaps with #45.

### LOST/REACQUIRE timeline

```
16:24:28  LOCK bowl
16:24:31  LOST → 4s search  → REACQUIRE bowl (16 frames, sim=0.941) ✓
16:24:39  LOST → 1s search  → REACQUIRE "cup" (10 frames, sim=0.714) — same bowl, label flicker
16:24:48  LOST → 9s search  → REACQUIRE "banana" (73 frames!, sim=0.825) ← embedding gate blocked
16:25:16  LOST → 2s search  → REACQUIRE "cup" (18 frames, sim=0.846)
16:25:22  LOST → 2s search  → REACQUIRE bowl (5 frames, SIZE OVERRIDE 8.337x, sim=0.782)
16:25:27  LOST → 3s search  → REACQUIRE bowl (12 frames, sim=0.817)
16:25:32  LOST → 2s search  → REACQUIRE "banana" (13 frames, sim=0.945)
16:26:04  LOST → user CLEAR
```

### VT survival per lock

| # | Duration | Bowl visible at LOST? | Cause |
|---|----------|----------------------|-------|
| 1 | 3s | No — panned to floor | Left frame |
| 2 | 4s | Yes — at edge | Rotation |
| 3 | 8s | Barely — sliver at edge | Rotation |
| 4 | 10s | No — wall/ceiling | Full rotation |
| 5 | 6s | No — white wall | Rotation |
| 6 | 4s | **Yes — centered!** | VT died on visible object |
| 7 | 3s | No — floor/cabinet | Left frame |
| 8 | 2s | Barely — edge | Rotation |
| 9 | 30s | **Yes — zoomed out** | Zoom-out made bowl tiny |

## Fix Requirements

The fix must:
- Allow reacquisition to work during rotation (when embeddings can't bridge frames)
- NOT re-introduce cross-category false reacquisitions (the reason the embedding gate exists)
- NOT be parameter tweaking — needs a structural solution

## Related

- #45 — VT confidence / tracking responsiveness
- PR #52 — Identity improvements that added the embedding gate
- #43 — Original tracking responsiveness (closed)

tuky191:
## Summary

Three structural fixes to tracking, bundled because they live in the same code path and were discovered iteratively:

1. **Template self-verification for VT drift** (#45) — replaces detector-dependent drift detection as the primary signal. Every 5 frames during VT tracking, embed the current crop and compare to the lock gallery. 3 consecutive similarities below 0.4 (~0.5s) triggers drift. Detector cross-check remains as a secondary safety net.

2. **Sync embedding fallback during rotation** (#53) — when the async embedding cache can't merge across frames (IoU < 0.3 due to position shift), synchronously embed the closest same-category candidate (~23ms one-time cost). Fixes 9-second reacquisition gaps where correct detections were hard-rejected because the embedding gate couldn't verify identity.

3. **Filter phantom detections before scoring** (#53) — `DetectionFilter` now runs before `ReacquisitionEngine.processFrame()` and `maxBoxArea` is lowered from 0.7 to 0.5. Prevents full-screen phantom detections (airplane/book at ~70% frame area from motion blur during rotation) from reaching the scoring pipeline.

## Research basis (#45)

The old VT drift detection had a structural flaw: it treated "detector found something elsewhere in the frame but not at VT position" as drift evidence, which is wrong — it's detector absence, not tracker drift. This caused false drift kills on small/uniform objects (yellow bowls, etc.) that EfficientDet can't reliably see.

Three attempts at fixing the detector cross-check (IoU > 0.05 ring, three-way signal classification with driftScore, DeepSORT-style max_age cap) all ran into parameter-tuning problems. Research on mature trackers (TLD PAMI 2012, Stark ICCV 2021, DaSiamRPN ECCV 2018, DeepSORT, ByteTrack, OC-SORT) consistently shows the primary drift signal should be the tracker's own self-assessment, not the detector. Template self-verification is the direct implementation of that principle.

## Device-tested behavior

- Flowerpot scenario: 11/12 reacquisitions within 2-22 frames, no false drift kills
- Template mismatch fired once correctly (sim=0.32) right before VT confidence also collapsed
- Remaining issues (e.g., 12-second gaps during 180° rotation) are detector coverage problems, not drift

## Files changed

| File | Changes |
|------|---------|
| `ObjectTracker.kt` | +84 lines: sync embedding fallback (~40 lines), template self-verification (~20 lines), filter-before-scoring (~4 lines), reset state in clear/reacquire paths |
| `DetectionFilter.kt` | `maxBoxArea` 0.7 → 0.5 |
| `CLAUDE.md` | Document all three changes + research rationale + new tunables + logcat tags |

## Test plan

- [x] 225+ unit tests pass
- [x] Device-tested manually across several sessions (bowl, flowerpot, toilet, laptop, mouse)
- [ ] On-device video replay tests — deferred pending code review
- [ ] Code review before merge

## Related

- Closes #45 (VT drift responsiveness — R3 from issue plan)
- Closes #53 (embedding delivery during rotation + phantom detections)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -46,7 +46,7 @@ app/src/main/java/com/haptictrack/
 │   ├── PersonAttributeClassifier.kt         # Crossroad-0230 + BlazeFace + age-gender
 │   ├── VisualTracker.kt                     # OpenCV VitTracker wrapper
 │   ├── FrameToFrameTracker.kt               # IoU-based detection ID assignment
-│   ├── DetectionFilter.kt                   # Noise removal (confidence, size, aspect ratio)
+│   ├── DetectionFilter.kt                   # Noise removal (confidence, size, aspect ratio; applied before scoring)
 │   ├── ScenarioRecorder.kt                  # JSON capture for replay testing + serialization helpers
 │   ├── DebugFrameCapture.kt                 # Annotated PNGs + session.log on tracking events
 │   ├── ObjectSegmenter.kt                   # magic_touch segmentation for masked crops
@@ -94,16 +94,18 @@ CameraViewModel
 2. `DeviceOrientationListener` provides physical rotation → bitmap rotated to upright
 3. `VisualTracker` (VitTracker) updates the locked object's position by pixel correlation
 4. Detector runs in parallel to cross-check — if detector confirms the label at the tracker's box, tracking continues
-5. If detector doesn't confirm for 10 frames → drift detected, visual tracker stopped
+5. Every 5 frames, the current VT crop is embedded and compared against the lock gallery — primary drift signal (catches drift even when the detector can't see the target)
+6. 3 consecutive template mismatches (sim < 0.4) OR 10 frames without detector confirmation → drift detected, visual tracker stopped
 
 **Re-acquisition (object lost):**
 1. Detector runs, `FrameToFrameTracker` assigns stable IDs via IoU matching
 2. `Yolov8Detector` enriches COCO labels with OIV7 labels (every 10th search frame)
-3. `AppearanceEmbedder` computes visual fingerprints for all candidates
-4. `ReacquisitionEngine` scores candidates: position (decays over time) + size + label (20% scoring factor, -0.5 penalty for mismatch) + appearance similarity + color histogram + person attributes
-5. Strong embedding match (>0.7 cosine similarity) overrides position/size hard thresholds
-6. Best candidate above `minScoreThreshold` becomes the new lock; visual tracker re-initializes
-7. `ScenarioRecorder` captures every frame's detections and events as JSON for replay testing
+3. `AppearanceEmbedder` computes visual fingerprints for all candidates (async pipeline; synchronous fallback for the closest same-category candidate when the async cache can't bridge across frames during rotation)
+4. `DetectionFilter` runs before scoring to strip phantom full-screen detections (area > 0.5) that EfficientDet produces during motion blur
+5. `ReacquisitionEngine` scores candidates: position (decays over time) + size + label (20% scoring factor, -0.5 penalty for mismatch) + appearance similarity + color histogram + person attributes
+6. Strong embedding match (>0.7 cosine similarity) overrides position/size hard thresholds
+7. Best candidate above `minScoreThreshold` becomes the new lock; visual tracker re-initializes
+8. `ScenarioRecorder` captures every frame's detections and events as JSON for replay testing
 
 **Two-stage detection:**
 - EfficientDet-Lite2 (COCO 80) runs every frame for reliable bounding boxes
@@ -274,6 +276,9 @@ Position (50%, decays) + size (25%) + base bonus (25%). All survivors already pa
 ## Current Work (may be stale — verify with git/GitHub)
 
 **Recently merged to master:**
+- PR #54 — Unify on SurfaceTexture pipeline, remove 3-stream mode (#48)
+- PR #52 — Identity: tentative confirmation, ratio test, gallery-relative threshold, online classifier (#50)
+- PR #44 — Velocity estimation + async embeddings + SurfaceTexture recording
 - PR #42 — GPU delegate for all models, FP16 detector, adaptive frame skip, stealth mode, dependency bumps
 - PR #40 — Regression baseline scenarios with quantitative thresholds
 - PR #37 — Manual camera controls: pinch zoom, 4K recording, stealth mode, volume-down hands-free cycle
@@ -295,8 +300,15 @@ Position (50%, decays) + size (25%) + base bonus (25%). All survivors already pa
 ### Two-stage detection: EfficientDet-Lite2 + YOLOv8n-oiv7 (decided 2026-04-17)
 EfficientDet-Lite2 runs every frame for reliable bounding boxes (person at 0.80-0.88 confidence). YOLOv8n-oiv7 runs on-demand to upgrade coarse COCO labels to finer OIV7 labels (601 classes). OIV7 splits "person" across sub-classes (Boy, Girl, Man, Woman) so no single class gets high confidence indoors — that's why YOLOv8 can't be the sole detector. Enrichment is semantically guarded: COCO "person" can only become Man/Woman/Boy/Girl/Human face, never furniture.
 
-### Two-tier tracking: VisualTracker + Detector (decided 2026-04-13)
-VitTracker (OpenCV) follows the locked object by pixel correlation — no classifier needed, very stable frame-to-frame. But it drifts when the object leaves frame. The detector + embedding pipeline handles re-acquisition. The visual tracker is cross-checked against the detector every frame; 10 consecutive unconfirmed frames triggers drift detection and handoff to re-acquisition.
+### Two-tier tracking: VisualTracker + Detector (decided 2026-04-13, revised 2026-04-24)
+VitTracker (OpenCV) follows the locked object by pixel correlation — no classifier needed, very stable frame-to-frame. But it drifts when the object leaves frame. The detector + embedding pipeline handles re-acquisition.
+
+**Drift detection uses two parallel signals:**
+1. **Template self-verification (primary)**: every 5 frames during VT tracking, embed the current VT crop (raw crop fallback, ~8ms) and compare to the lock gallery via `bestGallerySimilarity()`. 3 consecutive similarities below 0.4 (~0.5s) triggers drift. Detector-independent — catches drift even when the detector can't see the target (small objects, uniform surfaces, label flicker).
+2. **Detector cross-check (secondary)**: if the detector doesn't confirm the VT position for 10 consecutive frames, also triggers drift. Kept as a safety net for cases where the template somehow keeps matching but the object has moved.
+
+### Template self-verification rationale (decided 2026-04-24)
+The old approach (detector cross-check only) had a structural flaw: "detector found something elsewhere in the frame but not at VT position" was counted as drift evidence, even when the detector simply missed our object (common for small/uniform objects like a yellow bowl). This caused false drift kills on correctly-tracking VT. Research (TLD PAMI 2012, Stark ICCV 2021, DaSiamRPN ECCV 2018) consistently treats the tracker's own self-assessment as the primary drift signal, with the detector as a peer rather than a judge. Template self-verification directly measures "does this crop still look like the locked object" — the right question.
 
 ### COCO label stored alongside enriched label (decided 2026-04-17)
 `lockedCocoLabel` is stored at lock time alongside the enriched label. Label matching accepts either: a candidate labeled "potted plant" (COCO) matches a lock of "flowerpot" (OIV7) because the COCO parent is "potted plant". Without this, enrichment-timing mismatches cause the correct object to get the -0.5 label penalty.
@@ -360,11 +372,12 @@ When VitTracker is confident (>0.6) and has 5+ confirmed frames with 0 unconfirm
 | Tag | What it logs |
 |---|---|
 | `Reacq` | LOCK, LOST, SEARCH (with candidate scores + similarity), REACQUIRE (with hop count + sim), TIMEOUT, CLEAR, GIVE_UP, OVERRIDE (when embedding bypasses geometric filters) |
-| `VisualTracker` | INIT (tracker started), LOST (confidence dropped), DRIFT (unconfirmed frames exceeded threshold) |
+| `VisualTracker` | INIT (tracker started), LOST (confidence dropped), DRIFT (template mismatch OR unconfirmed frames exceeded), Template mismatch (per-check log when sim below threshold) |
 | `FTFTracker` | NEW (fresh ID assigned), MATCH (IoU match to previous frame) |
 | `Yolov8Det` | Enriched label results, no-enrichment cases |
 | `ScenarioRec` | Recording started/stopped, frame count |
-| `AppearEmbed` | Embedding failures |
+| `AppearEmbed` | Embedding failures, gallery additions |
+| `EmbedSync` | Synchronous embedding fallback fired for a candidate that missed the async cache |
 | `TFLiteGPU` | GPU delegate activation or CPU fallback per model |
 | `DebugCapture` | Saved debug frame filenames |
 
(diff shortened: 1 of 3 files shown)


tuky191 (commented):
## Code Review: PR #55 — Fix VT drift detection + embedding delivery (#45, #53)

225 unit tests pass. +120 −19, 3 files, 3 commits. Three structural fixes bundled because they share the same code path.

### No high-severity issues

### Medium

**1. Template verification has no hysteresis — single threshold is a cliff** (`ObjectTracker.kt:335-339`)

The template check uses a single threshold (`TEMPLATE_SIM_THRESHOLD = 0.4`): sim below → increment counter, sim at or above → reset counter to 0. This creates a cliff edge: if the object oscillates near 0.4 (e.g., 0.38, 0.42, 0.39), the counter resets on every "good" frame and drift is never detected.

Consider a two-threshold hysteresis: increment below 0.35, reset only above 0.45, hold the counter in between. This prevents marginal oscillation from masking real drift.

Not urgent — `TEMPLATE_MISMATCH_MAX = 3` consecutive frames at the same check rate means this would need sustained oscillation across ~0.5s to matter, which is unlikely in practice.

**2. Sync fallback picks candidate by position proximity** (`ObjectTracker.kt:493-498`)

During rotation, the locked object can appear anywhere in the frame. The sync fallback picks the closest same-category candidate to `lastKnownBox` to embed. During heavy rotation, the correct object may not be the closest one — `lastKnownBox` is stale (position decay helps in scoring but not in candidate selection here).

Impact is limited: worst case is one wasted ~23ms embedding on the wrong candidate. The correct candidate gets async embeddings next frame. But worth noting — if sync fallback consistently picks the wrong candidate during rotation, it burns 23ms/frame for no gain.

**3. `maxBoxArea` 0.5 may reject legitimate large-subject detections** (`DetectionFilter.kt:13`)

0.7 → 0.5 rejects any detection covering >50% of the frame. A person filling the viewfinder during close-up tracking could exceed this. The filter runs before scoring now, so such a detection would be invisible to `ReacquisitionEngine`.

Did you see any legitimate detections in the 0.5–0.7 range during testing? If so, consider whether this should be scoped to search-only (not during VT confirmation path). Currently the VT confirmation path uses `tracked` before filtering, so it's not affected — but worth confirming.

### Low

**4. Debug capture still uses unfiltered `withEmbeddings`** (`ObjectTracker.kt:565`)

`captureDebugFrame(bitmap, withEmbeddings, ...)` passes unfiltered detections while `processFrame` receives `filtered`. This is correct for diagnostics (you want to see what the filter removed), but the SEARCH debug log will show candidates that `processFrame` never scored. Could be confusing when debugging a specific scenario. Consider a comment noting this is intentional.

**5. Template check runs on `vtFrameCounter % 5` but only on non-skip frames** (`ObjectTracker.kt:329-331`)

When VT frame skipping is active (`skipDetector=true` on every other frame), the template check condition `!skipDetector && vtFrameCounter % TEMPLATE_CHECK_INTERVAL == 0` means it runs every ~10 frames instead of every 5 — roughly every 0.33s at 30fps. This is fine for drift detection timing, but the effective check rate is implicitly coupled to the skip interval. A comment noting "effectively every ~10 frames when skip is active" would help future readers.

- **Template self-verification is well-placed in the code flow**: runs after detector confirmation/contradiction logic and scene negative collection, before the drift decision. This means the template check doesn't interfere with the detector cross-check — both signals contribute independently to the drift decision.

- **State resets are comprehensive**: `templateMismatchCount` is reset in all 5 relevant paths — `clearLock()`, counter >= threshold reset, drift kill, VT lost, and re-acquire. No leakage across tracking sessions.

- **No coordinate bugs in the unified SurfaceTexture pipeline**: With `deviceRotation = 0` always (PR #54), `unmapRotation` is identity, so `rawBox == vtBox` and all crop coordinates are in display space. The rotation mismatch bugs from PR #52 can't recur.

- **Scenario recorder captures filtered detections**: `scenarioRecorder.recordFrame(filtered)` — replays will match what `processFrame` actually scored. Correct.

- **Research basis is solid**: TLD/Stark/DaSiamRPN all use tracker self-assessment as primary drift signal. The implementation here (gallery comparison via cosine similarity) is a clean adaptation of the concept.

Good to merge. The hysteresis suggestion (#1) is worth considering for #56 if it's stacked on top.

PR #56, opened 2026-04-24
Detector rotation + VT size/hysteresis + volume-up stealth

tuky191:
Stacked on top of #55. Three independent improvements:

## 1. Detector rotation fix

EfficientDet fails on rotated scenes — when the phone is physically rotated (deviceRotation != 0), the bitmap from the SurfaceTexture GL pipeline contains a rotated scene and the detector produces 0 candidates despite the target being visible.

`runDetector` now reads the physical rotation from `deviceRotationProvider`, pre-rotates the bitmap to physical-upright before inference, and uses `unmapRotation` to convert output boxes back to display-space for the rest of the pipeline.

Fixes 12-second reacquisition gaps during rotation that were observed in the flowerpot session.

## 2. VT drift: box size check + template hysteresis

Two additional drift signals to catch the "VT expands to whole screen" failure mode that was slipping past the existing signals:

- **Box size sanity**: if VT's output box area exceeds 5x the area of the box it was initialized with, kill immediately. Objects don't grow 5x in a fraction of a second — this means VT latched onto background texture.
- **Template mismatch hysteresis**: replaces single 0.4 threshold with asymmetric thresholds. Increment counter below 0.4, reset only above 0.5. The 0.4–0.5 dead zone holds the counter so oscillation at the boundary can't mask real drift.

Both signals verified on device: fire within 1–2s of a bad reacquire (previously ~10s via the detector cross-check safety net).

## 3. Stealth mode UX: volume-up control, touch blocked

- Volume-up now toggles stealth mode (mirrors the volume-down lock/record cycle). Volume-down behavior unchanged.
- Removed the on-screen STEALTH button.
- Swallow all touches on the preview while in stealth mode. Volume-up is the only way in/out — prevents accidental taps from exiting stealth or locking on objects.

## Test plan

- [x] 225+ unit tests pass
- [x] Device-tested manually across multiple sessions (flowerpot, dog, mouse, tv)
- [x] Size-drift signal fired correctly on 2 real "expanded to whole screen" events
- [x] Template hysteresis resets only on clear recovery (sim > 0.5)
- [x] Stealth mode: volume-up toggles, touches blocked while in stealth

## Related

- Builds on #55
- Closes remaining sub-items of #45

### DIFF
--- a/app/src/main/java/com/haptictrack/MainActivity.kt
+++ b/app/src/main/java/com/haptictrack/MainActivity.kt
@@ -29,6 +29,10 @@ class MainActivity : ComponentActivity() {
             viewModel.onVolumeDown()
             return true
         }
+        if (keyCode == KeyEvent.KEYCODE_VOLUME_UP) {
+            viewModel.onVolumeUp()
+            return true
+        }
         return super.onKeyDown(keyCode, event)
     }
 }
--- a/app/src/main/java/com/haptictrack/tracking/ObjectTracker.kt
+++ b/app/src/main/java/com/haptictrack/tracking/ObjectTracker.kt
@@ -208,9 +208,11 @@ class ObjectTracker(
     }
 
     /**
-     * Process a display-oriented bitmap from PreviewView.getBitmap().
-     * Unlike the ImageProxy path, the bitmap is already in screen orientation —
-     * no rotation is applied, and detector coordinates are used as-is (deviceRot=0).
+     * Process a display-oriented bitmap from the SurfaceTexture GL pipeline.
+     * The bitmap pixels are in display (phone-top) orientation. The detector
+     * reads [deviceRotationProvider] internally and rotates the bitmap to
+     * physical-upright before inference; VT and the rest of the pipeline
+     * continue operating in display-image coordinates.
      */
     fun processBitmap(bitmap: Bitmap) {
         processBitmapInternal(bitmap, deviceRotation = 0)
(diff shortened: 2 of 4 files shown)


tuky191 (commented):
## Code Review: PR #56 — Detector rotation + VT size/hysteresis + volume-up stealth

225 tests pass. +93 −48, 4 files, 3 commits. Stacked on #55. Three independent changes.

### Addresses #55 review items

- **#55 finding #1 (template hysteresis)** — Directly addressed. Single threshold replaced with asymmetric `TEMPLATE_SIM_LOW=0.4` / `TEMPLATE_SIM_OK=0.5` and dead zone in between. Clean `when` block. Exactly what was suggested.

### Medium

**1. `deviceRot` parameter is dead in `runDetector`** (`ObjectTracker.kt:637`)

`runDetector` still accepts `deviceRot: Int` but never reads it — it uses `actualRot` from `deviceRotationProvider` instead. All call sites pass `deviceRotation` which is always 0 in the unified SurfaceTexture pipeline. The parameter should be removed (and `frameWidth`/`frameHeight` too — they're superseded by `uprightW`/`uprightH`).

Not a bug, just dead parameters that will confuse future readers into thinking rotation is handled by the caller.

**2. Bitmap rotation allocates on every non-portrait frame** (`ObjectTracker.kt:644, 672-674`)

`rotateBitmap()` creates a new `Bitmap` via `Matrix.postRotate` on every detector frame when rotation != 0. At 640×854 ARGB that's ~2.1MB per frame × 10-12fps = ~21-25MB/s of allocations. The bitmap is recycled after use (line 668), but this is still significant GC pressure.

Not a regression (the old ImageProxy path did the same rotation), but now that rotation always goes through this path, consider pre-allocating a reusable bitmap or using the GL pipeline to handle rotation (SurfaceTexture already has transform matrices). Low priority — the GC handles it at this scale, and the rotation path is only active when the phone is physically rotated.

### Low

**3. VT box area check doesn't account for zoom changes** (`ObjectTracker.kt:373-374`)

`vtLockedBoxArea` is set at lock/reacquire time. Auto-zoom adjusts the camera zoom ratio during tracking, which changes the apparent object size. A 2.24x zoom increase would push area to 5x, triggering a false drift kill. In practice the zoom speed (0.10/frame) and the 0.15 target occupancy make this unlikely — the object would need to go from 0.15 to 0.75 area, requiring massive zoom change. But worth noting if zoom behavior changes later.

**4. Volume keys are now fully consumed** (`MainActivity.kt:32-35`)

Both volume-up and volume-down return `true` (consumed), suppressing system volume adjustments while the app is active. This is the existing pattern from volume-down, so consistent. Users may expect volume keys to work normally when not in stealth/tracking — but this is a UX decision, not a code issue.

- **Detector rotation logic is correct**: `-actualRot` counteracts the phone's physical rotation, producing an upright bitmap for inference. `unmapRotation(actualRot)` correctly maps detector output back to display space. Traced through all 4 rotation cases (0/90/180/270).

- **Template hysteresis dead zone**: `when` block cleanly separates the three states — increment below 0.4, hold in [0.4, 0.5], reset above 0.5. Counter survives marginal oscillation. Good.

- **VT box size sanity**: 5x ratio is well-chosen — objects don't grow 5x during normal tracking, but VT's "expand to whole screen" drift produces 6-7x easily. Immediate kill (no consecutive-frame requirement) is correct since box expansion is catastrophic and unambiguous.

- **Stealth UX**: Touch swallow at the `pointerInteropFilter` level is the right place — blocks pinch-zoom, tap-to-lock, and all other gestures. STEALTH button removal and tap-to-exit removal are clean (−27 lines).

- **State management**: `vtLockedBoxArea` reset in all 5 paths (init 0, clearLock, drift kill, VT lost, reacquire set). Consistent with `templateMismatchCount` pattern from #55.

Good to merge (after #55 lands).

Issue #49, opened 2026-04-21
GL pipeline optimization: PBOs, bitmap pool, eliminate copies for +5-8fps


tuky191:
## Problem

SurfaceTexture processing runs at 14fps. An additional 5-10fps would significantly improve VT tracking stability (less displacement between frames = higher confidence = fewer losses).

### GL thread (~34ms budget, limited by camera delivery at 29fps)

| Step | Time | Notes |
|------|------|-------|
| updateTexImage + getTransformMatrix | ~1ms | |
| Render quad to FBO | ~1ms | |
| **glReadPixels** | **~8-12ms** | GPU→CPU sync stall, biggest bottleneck |
| Bitmap.createBitmap + copyPixelsFromBuffer | ~3-5ms | Allocation + pixel copy |
| **bitmap.copy for viewfinder** | **~3-5ms** | Full bitmap copy every frame |

### Processing thread (~71ms per frame = 14fps)

| Step | Time | Notes |
|------|------|-------|
| VT update (bitmapToMat + inference) | ~5ms | Converts ARGB→RGBA→BGR |
| **EfficientDet detector** | **~35-40ms** | GPU via MediaPipe, the big one |
| FTF assignment + scoring | ~1ms | |
| Async embeddings | off-thread | ~8ms per crop |

### 1. PBO (Pixel Buffer Objects) for async glReadPixels — saves ~5-8ms

The biggest single win. Currently `glReadPixels` blocks until GPU→CPU transfer completes. With PBOs:
- Frame N: start async read into PBO
- Frame N+1: map previous PBO (transfer already complete), start new read
- Adds 1 frame of latency but eliminates the GPU sync stall

```kotlin
// Pseudocode
GLES30.glBindBuffer(GLES30.GL_PIXEL_PACK_BUFFER, pbo[currentIndex])
GLES30.glReadPixels(0, 0, w, h, GL_RGBA, GL_UNSIGNED_BYTE, 0)  // async, returns immediately
// ... next frame, map the previous PBO
GLES30.glBindBuffer(GLES30.GL_PIXEL_PACK_BUFFER, pbo[previousIndex])
val buf = GLES30.glMapBufferRange(...)  // data is already there
```

Requires GLES 3.0 (available on all devices we target). Two PBOs ping-ponged.

### 2. Pre-allocated bitmap pool — saves ~3-5ms

[...]

Issue #35, opened 2026-04-17
Auto-lock: lock on predefined criteria (label + attributes)


tuky191:
## Feature

Auto-lock on a predefined target without manual tap. User selects criteria from UI, app automatically locks onto the first matching detection.

**Examples:**
- "Lock on any person"
- "Lock on men in their 40s"  
- "Lock on heavy-set man with backpack"
- "Lock on dog"

### Currently in the app
- **Label**: 80 COCO classes + 601 OIV7 classes via YOLOv8 enrichment
- **Gender**: male/female (Crossroad-0230 body + age-gender-retail-0013 face, 95.8% accuracy)
- **Age**: estimated from face (~7 year error, only when face visible)
- **Clothing**: long sleeves, pants, jacket, backpack, hat, bag (Crossroad-0230)
- **Clothing color**: upper/lower body dominant color (HSV quantized to 11 basic colors)

### Planned (#34)
- **Face embedding** (MobileFaceNet, ~1 MB) — identity, not attributes
- **Body re-ID embedding** (OSNet, ~2.2 MB) — identity, not attributes

### Missing for richer auto-lock
- **Body type/build**: no model currently. Options researched:
  - **RAP v2 dataset** has body build labels (thin/normal/fat, 84K images). Train MobileNetV3-Small → ~4-5 MB model, ~80-85% accuracy
  - **Pose landmark ratios** from MediaPipe Pose (shoulder-to-hip ratio). Zero model cost but viewpoint-sensitive and coarse
  - **Silhouette contour analysis** using existing `magic_touch` segmenter — extract width profile at shoulder/waist/hip. No new model, ~10 fps
- **Race/ethnicity**: models exist (FairFace ~7 MB) but app store / legal risk (GDPR, BIPA). Not recommended.

## UI concept

Settings screen or overlay with filter selectors:
- Label dropdown (person, dog, car, cup, ...)
- For "person": gender toggle, age range slider, body type selector, clothing filters
- "Start auto-lock" button → app scans detections each frame, locks on first match

## Dependencies

- #34 — Person identity (face + body re-ID) should land first for reliable person tracking after auto-lock
- Settings UI — currently no settings screen exists

tuky191:
## Update: Model stack changes affecting auto-lock signals (2026-04-30)

### Crossroad-0230 dropped (#124)

[...]

PR #60, opened 2026-04-24
Processing-thread perf: pool + retention + async PNG + rotated pool + drop YOLOv8 (#49)

tuky191:
## Summary

Performance work for #49, measured on a Xiaomi Poco F4 (Snapdragon 870). Split into four logical commits on this branch.

**Headline:** active-tracking fps moved from 6-17fps (40-158ms/frame) on master to a **sustained 22-24fps (41-45ms/frame)** — floor close to the detector's inherent ~35ms cost. Reacquire latency dropped from ~2.5s to ~1-2s.

## Instrumentation first

Neither thread was instrumented. First commit adds timing logs under tag `STFrameReader`:
- `GL frame avg: readPx / copy / total` every 60 GL frames
- `Process frame avg: XXms (~YYfps)` every 60 processed frames

Measurements revealed the GL thread (3-6ms/frame) was nowhere near the 33ms camera budget — the real bottleneck was the processing thread. That shifted the optimization target away from the issue's original Phase 2 (PBO async readback) toward the per-frame cost of `ObjectTracker.processBitmap`.

### 1. GL pipeline: pool output + viewfinder bitmaps

`SurfaceTextureFrameReader` was allocating a fresh ~2MB bitmap every GL frame and another via `bitmap.copy()` for the viewfinder — ~58MB/s of allocation pressure.

- Two 3-entry `BitmapRing` pools (processing + viewfinder)
- Viewfinder path uses a second `copyPixelsFromBuffer` from the same `readBuffer` — no more `.copy()`
- 3-entry in-flight queue ensures Compose never redraws a bitmap being overwritten
- Ownership contract: `ObjectTracker.processBitmap` no longer recycles its input — the caller (pool in production, decoder in tests) owns lifetime

### 2. Retain input bitmap as lastFrameBitmap, kill per-frame copy

`processBitmap` was calling `bitmap.copy()` (~4ms, 2MB memcpy) every frame just to keep a reference for `lockOnObject` and debug capture. Since the input is a stable snapshot for the duration of the call, we can retain it directly and release the previous `lastFrameBitmap` back to the pool via a new `bitmapRecycler` callback.

### 3. Remove YOLOv8n-oiv7 label enrichment

After the binary person/not-person gate landed in #52, YOLOv8 label enrichment stopped contributing to any gate or score — it only changed the display label ("person"→"woman", "potted plant"→"flowerpot"). Removing it saves:

- ~270ms at lock time
- ~1-2s at app startup (no "Loading YOLOv8 (GPU)")
- 6.8MB APK size
- One GPU interpreter

For auto-lock (#35), `PersonAttributeClassifier` already provides gender/clothing/age as structured fields — a better representation than OIV7 labels anyway.

### 4. Flush debug PNGs async; pool rotated bitmap

Two final cuts on the processing thread:

- **Async PNG encoding:** `DebugFrameCapture.capture()` snapshots the bitmap (~4ms) and submits to a dedicated single-thread `DebugCapture-IO` executor with bounded queue + DiscardOldestPolicy. Per-event cost: ~50-100ms → ~4ms.
- **Rotated bitmap pool:** `runDetector` reuses a single `uprightBuffer` drawn into via `Canvas`, reallocating only when target dimensions change. Per-frame cost when phone tilted: ~5-10ms → ~1-2ms.

## Measurements

All on-device, active tracking on a mouse with rotation:

| State | Active fps | Per-frame ms (typical) | Per-frame ms (spikes) |
|---|---|---|---|
| master | 6-17fps | 60-80ms | 107-158ms |
| +GL pool +retention | 10-22fps | 40-60ms | 98-122ms |
| **+YOLOv8 removed +async PNG +rotated pool** | **22-24fps** | **41-45ms** | **60-80ms** |

The floor of ~44ms is now detector-bound (~35ms EfficientDet + ~10ms misc). Next structural wins would be on the detector side (frame skip tuning, smaller model) — separate work.

## Test plan

- [x] `./gradlew testDebugUnitTest` — all 184 unit tests pass
- [x] `chair_living_room_tracking_rate` device replay test passes
- [x] Live device session: lock on mouse, survive rotation, reacquire correctly
- [ ] Soak test 5+ min for bitmap leaks / pool exhaustion (recommended before merge)

## Follow-up (separate PR)

GL PBO async readback (original Phase 2): save ~2-3ms on GL thread. Now much lower priority since GL thread is not the bottleneck; will open a separate PR if the micro-win matters.


tuky191 (commented):
## Code Review: PR #60 — GL pipeline: pool output + viewfinder bitmaps (#49)

226 tests pass. +144 −22, 6 files. Replaces per-frame bitmap allocations with pooled rings.

### Medium

**1. `BitmapRing.release()` doesn't enforce size cap — comment/code mismatch** (`BitmapRing.kt:35-42`)

Comment says "Drops it if the pool is already full" but the code unconditionally adds:
```kotlin
synchronized(free) { free.addLast(bitmap) }
```
No size check. If `acquire()` creates overflow bitmaps during back-pressure and those are later released, the pool grows beyond `size`. In practice this is bounded by transient back-pressure (1-2 extra bitmaps), but the code should match the documented contract. Fix:
```kotlin
synchronized(free) {
    if (free.size < maxSize) free.addLast(bitmap)
    else bitmap.recycle()
}
```
Store the initial `size` as `private val maxSize`.

### Low

**2. Viewfinder in-flight queue relies on timing assumption** (`SurfaceTextureFrameReader.kt:171-178`)

The 3-entry `inflightViewfinder` queue assumes Compose's RenderThread finishes drawing within 2 frames (~67ms at 29fps GL rate). If RenderThread falls behind (heavy GC, GPU thermal throttle), a bitmap could be returned to the pool and overwritten while Compose is still reading it. Result: visual tearing, not a crash.

The previous approach (never recycle, let GC handle) was safe but generated ~35MB/s churn. The pool approach is the right tradeoff — just noting the assumption. A 4-entry pool would add ~2MB safety margin if tearing is observed.

**3. `processBitmapInternal` still copies bitmap to `lastFrameBitmap`** (`ObjectTracker.kt:~493, ~380`)

Both the VT and detector paths do `lastFrameBitmap = bitmap.copy(...)`. This ~2MB copy per frame survives from the pre-pool era. Now that the input bitmap has a predictable lifetime (released after `processBitmap` returns), you could potentially hold a reference to it instead of copying — but that would require extending the pool lease, which complicates the ownership model. Not worth changing now; just noting it's the next allocation to target per the PR description's "Phase 2" note.

- **Bitmap ownership contract is clean**: GL thread produces → `latestFrame` AtomicReference → processing thread consumes → `onFrame` → `processBitmap` → CameraViewModel `finally` releases to pool. Exception safety is correct — the `try/finally` in CameraViewModel ensures release even if `processBitmap` throws.

- **Pool exhaustion is graceful**: `acquire()` falls back to fresh allocation instead of blocking. This is the right choice — a one-frame allocation spike is better than blocking the GL thread or processing thread.

- **Viewfinder double-read from `readBuffer`**: Two `copyPixelsFromBuffer` calls from the same readback buffer (one for processing, one for viewfinder) avoids the old `bitmap.copy()` and its ~6-10ms cost. The buffer rewind between copies is correct.

- **VideoReplayTest cleanup**: Tests now explicitly recycle bitmaps with comments explaining the ownership difference from production. Correct.

- **Frame-timing instrumentation**: Averaged over 60-frame windows, logs readPx/copy/total under `STFrameReader`. Useful for profiling and self-documents the GL thread cost.

Good to merge after fixing the `release()` size cap (#1).

tuky191 (commented):
## Follow-up review (4 new commits: 4548881..a51e174)

226 tests pass. 4 new commits add processing-thread FPS log, YOLOv8 removal, bitmap retention, async debug capture, and rotated bitmap reuse. Significant scope expansion from the original pool-only PR.

### Verification of original finding

**#1 (Med) BitmapRing.release() size cap** — **Still open.** Code unchanged; comment still says "Drops it if the pool is already full" but no size check exists. With bitmap retention, pool growth during `lockOnObject` is more pronounced: the lock is held ~200-300ms for embeddings, during which the GL thread overflows the pool with ~6-9 extra allocations that permanently grow the deque.

### New findings

#### Medium

**2. `onAnalysisFrame` callback lost exception safety** (`CameraViewModel.kt:121-126`)

Previous version:
```kotlin
try { tracker.processBitmap(bitmap) }
finally { cameraManager.releaseAnalysisBitmap(bitmap) }
```

New version:
```kotlin
if (isTrackerReady) {
    tracker.processBitmap(bitmap)
} else {
    cameraManager.releaseAnalysisBitmap(bitmap)
}
```

If `processBitmap` throws, the bitmap IS released (via the internal `finally` + `bitmapRecycler`), so no leak. But the uncaught exception kills the processing thread in `SurfaceTextureFrameReader`, which has no exception handler. Tracking silently freezes — the GL thread keeps producing frames into `latestFrame`, the old one gets returned to the pool each time, but nobody processes them.

This was pre-existing (the processing thread never had an exception handler), but the old `try/finally` in the callback at least kept the pool healthy. Consider wrapping `onFrame` in the processing thread with a try-catch that logs but continues.

#### Low

**3. `lockOnObject` holds a pool bitmap for ~200-300ms** (`ObjectTracker.kt:153-200`)

The entire `lockOnObject` runs under `synchronized(lastFrameLock)`, which now holds a pool bitmap (not a copy). During those ~200-300ms of embedding/segmentation/attribute work, the processing thread blocks (can't swap `lastFrameBitmap`), and the GL thread overflows the pool. Functionally correct, but the pool growth is worse than with the old copy-on-retain pattern.

Not a regression in behavior (the processing thread was always blocked during `lockOnObject`), just a side effect of the retention optimization.

- **YOLOv8 removal is clean**: `Yolov8Detector.kt` deleted, `yolov8n_oiv7.tflite` (7.2MB) and `oiv7_labels.txt` removed, all references eliminated. `lockOnObject` now passes `label` directly (no enrichment). `ScenarioReplayTest.PERSON_LABELS` retains OIV7 variants for backward compatibility with captured scenarios. Saves ~7.2MB APK size + ~270ms lock latency.

- **`releaseOnExit` pattern is correct**: Initially points to input bitmap. After retention, swaps to previous `lastFrameBitmap`. `finally` block releases whichever is appropriate. Exception before the swap → releases input. Exception after the swap → releases previous. No double-release or leak.

- **Async debug capture is well-designed**: Bounded queue (8 entries) with `DiscardOldestPolicy` prevents pipeline lag. `bitmap.copy()` happens synchronously to snapshot the frame, then Canvas/PNG/IO runs on a low-priority thread. `shutdown()` with `awaitTermination(2s)` added and wired from `ObjectTracker.shutdown()`.

- **`rotateIntoBuffer` is correct**: Canvas-based rotation into a reusable bitmap. Reallocates only on dimension change (90°↔180° transition). Matrix math: rotate around source center, then translate to align with target top-left. Recycled in `shutdown()`.

- **`runDetector` signature cleaned up**: Removed dead `frameWidth`, `frameHeight`, `deviceRot` parameters (flagged in PR #56 review). Now reads rotation internally from `deviceRotationProvider`.

- **Processing-thread FPS log**: Reports avg process time and FPS every 60 frames. Useful for profiling — complements the GL-thread timing log from the original commit.

Good to merge after fixing the BitmapRing size cap (#1). Finding #2 (processing thread exception handler) is pre-existing and can be addressed separately.

tuky191 (commented):
## Follow-up review (commit 9f7e894)

Both open findings addressed:

| # | Finding | Status |
|---|---------|--------|
| **1 (Med)** | BitmapRing.release() size cap | **Fixed.** `maxSize` stored, `release()` checks `free.size < maxSize` and recycles overflow bitmaps. Comment updated with rationale. |
| **2 (Med)** | Processing thread dies on exception | **Fixed.** `onFrame` wrapped in try-catch that logs and continues. Tracker's internal `finally` handles bitmap ownership; the processing loop stays alive. |

Good to merge.

PR #61, opened 2026-04-24
GL pipeline: PBO ping-pong async readback (#49)

tuky191:
## Summary

Final item on the #49 list. Upgrades the GL context to ES 3.0 and routes `glReadPixels` through two pixel buffer objects in a ping-pong pattern, so the GPU→CPU readback overlaps with the next frame's render work instead of blocking the GL thread.

## How it works

Each frame:
1. Draw external texture → FBO (unchanged)
2. `glBindBuffer(PIXEL_PACK_BUFFER, pbos[write])` + `glReadPixels(…, 0)` — **queues** the DMA readback; returns immediately.
3. `glBindBuffer(PIXEL_PACK_BUFFER, pbos[read])` + `glMapBufferRange` — maps the OTHER pbo, whose readback was issued on the previous frame. The driver has had a full frame to complete it, so mapping is near-zero wait.
4. `copyPixelsFromBuffer` from the mapped region into processing + viewfinder bitmaps.
5. `glUnmapBuffer` + swap write/read indices.

First frame is a no-op on the consumer side (nothing in the read pbo yet). `haveLastFrame` tracks this.

## Measurements (Snapdragon 870, live camera)

| Metric | Before (sync readPixels) | After (PBO) | Delta |
|---|---|---|---|
| `readPx` | 3-5ms | **0.04-0.07ms** | ~100× faster |
| `copy` | 1-2ms | 3-5ms | ~2× slower |
| `total` GL | 3-6ms | 2.6-5.7ms | **~0.5-1ms saved** |

The `readPx` call just queues the DMA now, so it's essentially free. The actual transfer cost surfaces when we `glMapBufferRange` + `copyPixelsFromBuffer` from a non-CPU-cached region. Net win on this hardware is modest because the original sync stall was already only ~3ms.

Processing thread is unchanged (~40-42ms / 20-24fps active) — PBO only affects the GL thread.

## Tradeoffs

- **+1 frame latency** (~33ms at 30fps) in the consumer path. Imperceptible in practice; tracking already has inherent latency from the detector (~35ms).
- **ES 3.0 required** — minSdk=29 so this is universally available; no fallback path needed.

## Why it's still worth landing on modern hardware

- Architectural: readback is no longer a hard GPU sync point. The GL thread can do other work (shader setup, texture updates) overlapping with DMA.
- Future-proofing: if we ever raise preview resolution (e.g. 4K readback), the sync stall scales with pixel count. PBO amortizes that.
- Small-but-real win today.

## Test plan

- [x] `./gradlew testDebugUnitTest` — 184 unit tests pass
- [x] Live camera: viewfinder renders, tracking works, no visible tearing
- [ ] Soak 5+ min for any leaks / driver issues (recommended before merge)
- [ ] Verify on a range of rotations (the 1-frame latency shouldn't matter but worth a check)

### DIFF
--- a/app/src/main/java/com/haptictrack/camera/SurfaceTextureFrameReader.kt
+++ b/app/src/main/java/com/haptictrack/camera/SurfaceTextureFrameReader.kt
@@ -80,9 +80,6 @@ class SurfaceTextureFrameReader(
     private var eglSurface: EGLSurface = EGL14.EGL_NO_SURFACE
     private var textureId: Int = 0
 
-    // Pre-allocated readback buffer
-    private lateinit var readBuffer: ByteBuffer
-
     /**
      * Start the GL thread and create the Surface.
      * Returns a [Surface] that should be provided to CameraX Preview via [Preview.SurfaceProvider].
@@ -107,8 +104,6 @@ class SurfaceTextureFrameReader(
                 surfaceLatch.countDown()
 
                 val bufferBytes = outputWidth * outputHeight * 4
-                readBuffer = ByteBuffer.allocateDirect(bufferBytes)
-                    .order(ByteOrder.nativeOrder())
 
                 // Create an FBO for offscreen rendering at output resolution
                 val fbo = IntArray(1)
--- a/app/src/main/java/com/haptictrack/camera/SurfaceTextureFrameReader.kt
+++ b/app/src/main/java/com/haptictrack/camera/SurfaceTextureFrameReader.kt
@@ -6,9 +6,11 @@ import android.opengl.EGL14
 import android.opengl.EGLConfig
 import android.opengl.EGLContext
 import android.opengl.EGLDisplay
+import android.opengl.EGLExt
 import android.opengl.EGLSurface
 import android.opengl.GLES11Ext
 import android.opengl.GLES20
+import android.opengl.GLES30
 import android.util.Log
 import android.util.Size
 import android.view.Surface
@@ -104,7 +106,8 @@ class SurfaceTextureFrameReader(
                 resultSurface = Surface(surfaceTexture)
                 surfaceLatch.countDown()
 
-                readBuffer = ByteBuffer.allocateDirect(outputWidth * outputHeight * 4)
+                val bufferBytes = outputWidth * outputHeight * 4
+                readBuffer = ByteBuffer.allocateDirect(bufferBytes)
                     .order(ByteOrder.nativeOrder())
 
                 // Create an FBO for offscreen rendering at output resolution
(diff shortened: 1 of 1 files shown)


tuky191 (commented):
## Code Review: PR #61 — GL pipeline: PBO ping-pong async readback (#49)

226 tests pass. +76 −30, 1 file. Clean PBO implementation.

### Low

**1. `readBuffer` is now dead** (`SurfaceTextureFrameReader.kt:84, 110`)

`readBuffer` (a `ByteBuffer.allocateDirect` of outputWidth×outputHeight×4 bytes, ~2MB) is still allocated at line 110 but never used in the GL loop — the PBO path maps its own `ByteBuffer` via `glMapBufferRange`. Remove the field and the allocation.

**2. `glMapBufferRange` null check logs but skips the frame silently** (`SurfaceTextureFrameReader.kt:194-196`)

If `glMapBufferRange` returns null (driver error), the frame is skipped — no bitmap goes into `latestFrame`, no viewfinder update. The `pboWrite`/`pboRead` swap still happens, so the next frame should recover. The log message is there, which is good. Just noting that a sustained mapping failure would freeze both tracking and the viewfinder with no visible indication to the user.

- **PBO ping-pong is textbook correct**: Frame N kicks off async readback into `pbos[pboWrite]`, then maps `pbos[pboRead]` which was initiated at frame N-1. Swap at end. `haveLastFrame` gates the first iteration correctly — nothing to map on the very first frame.

- **EGL context upgrade**: `EGL_OPENGL_ES2_BIT` → `EGL_OPENGL_ES3_BIT_KHR`, context version 2 → 3. Required for `GLES30.glMapBufferRange` and `GL_PIXEL_PACK_BUFFER`. minSdk=29 guarantees ES 3.0. No fallback needed.

- **PBO cleanup**: `GLES30.glDeleteBuffers(2, pbos, 0)` at line 246 in the normal cleanup path. The early-failure catch block at line 250 doesn't explicitly delete PBOs, but `releaseEGL()` destroys the entire context which implicitly frees them.

- **Buffer ordering**: `mapped.order(ByteOrder.nativeOrder())` at line 189. Correct — mapped buffers from the driver may not have the expected byte order without this.

- **GLES20 calls still used for drawing**: `drawQuad`, `glTexParameteri`, `glActiveTexture`, etc. are all GLES20 calls, which work fine in a GLES 3.0 context (backward compatible). Only the PBO calls use GLES30. Clean.

- **1-frame latency tradeoff**: Well-documented in code and PR description. At 30fps = 33ms additional latency, negligible compared to detector latency (~35ms) and the ~40ms processing pipeline.

Good to merge after removing the dead `readBuffer`.

tuky191 (commented):
## Follow-up review (commit 4598266)

Dead `readBuffer` field and allocation removed. Good to merge.

PR #63, opened 2026-04-24
VT skip interval: widen to 3 when long-stable (#62 A1)

tuky191:
## Summary
- Two-tier skip interval during visual tracking. **Base** (confidence >0.6, confirmed ≥5) keeps the existing every-2nd-frame detector cadence. **Stable** (confidence >0.7, confirmed ≥10) runs detector every 3rd frame (~67% skipped).
- Template self-verification (#45, every 5 frames) is the primary drift signal now, so detector cadence can be a slower safety net once VT has been stable for a while.
- First of three tier-1 follow-ups from #62.

## Measured impact (Snapdragon 870 / live camera, cup session)

| Phase | Master baseline | This branch |
|---|---|---|
| Pre-lock idle | ~30fps | ~30fps |
| Lock burst | 28ms spike | 28ms spike (unchanged — D1) |
| Active steady tracking | 41-45ms / 22-24fps | **39-42ms / 24-25fps** |
| Search / LOST spikes | 50-70ms | 50-70ms (unchanged — B2) |

~2-4ms saved per frame in cruise, ~1-2fps bump. Below the theoretical ~6ms because the stable gate isn't always satisfied during real tracking — VT wobbles in and out of stable as the subject moves. Correct behavior.

Session log confirms tracking quality unaffected: cup session reacquired 3× cleanly.

## Why this is safe
- The base regime (interval=2) is unchanged for the VT warm-up phase.
- Template self-verification runs every 5 frames regardless of detector cadence; it's independent of this change.
- Template mismatch / unconfirmed counters still trip drift detection at the same thresholds.
- Resets on drift, lock clear, or VT stop — exactly as before.

## Test plan
- [x] Unit tests pass (`./gradlew testDebugUnitTest`)
- [x] Build + deploy to device
- [x] Lock on a stable cup, observe `STFrameReader` log — active steady at 39-42ms, 24-25fps
- [x] Confirm reacquires still fire cleanly (3× in the test session)
- [ ] Video replay suite — run before merge

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #62, opened 2026-04-24
Follow-ups to #49: detector-side perf + unlocked GLES3 wins


tuky191:
## Context

After #49 (merged in #60 + #61), processing-thread tracking throughput doubled from 6-17fps to 22-24fps. The new floor (~44ms/frame) is detector-bound: ~35ms EfficientDet + ~9ms misc. The architectural work also put a bunch of previously-hard things within reach.

This issue tracks the next tier of optimizations, now unblocked.

## Remaining bottlenecks (from instrumented measurements)

After #60/#61 the typical active-tracking timeline per frame is roughly:

| Step | ms | Notes |
|---|---|---|
| EfficientDet-Lite2 (GPU) | ~35 | dominant |
| VT update (OpenCV VitTracker) | ~5 | CPU |
| Pool/callback overhead | ~1 | |
| Misc bookkeeping | ~3 | |
| **Typical total** | **~44** | |
| Sync embedding fallback (when fires) | +23 | only during search + rotation |
| YOLOv8 (removed in #60) | — | ✅ gone |
| Debug PNG encoding (moved off-thread in #60) | — | ✅ async |

### A. Detector-side — biggest remaining win

**A1. Widen `VT_SKIP_INTERVAL` when VT is highly confident**
Currently detector runs every 2nd frame when VT is confirmed. Raising to 3 when VT confidence > 0.7 AND has been confirmed ≥10 frames saves ~35ms on skipped frames. Risk: slower drift detection — template self-verification (#45) is the primary drift signal now, so detector cadence is a secondary safety net; slightly slower is tolerable.

**A2. Try EfficientDet-Lite0 instead of Lite2**
Lite0 is ~3× faster on GPU (~12ms vs ~35ms). Accuracy drops but the binary person/not-person gate is forgiving — we don't need the fine category distinctions. Worth A/B measuring on the full replay suite.

**A3. Parallelize detector with embedder on a separate thread**
Detector GPU inference and embedder GPU inference currently serialize on the processing thread. If the GPU has headroom (likely — one delegate at a time), running embedder async while detector runs cuts per-frame wall time when both fire. Requires care: both models use MediaPipe, which isn't documented as multi-GPU-session safe.

### B. Sync embedding fallback spikes (60-80ms, firing during search)

**B1. Replace "one best candidate per frame" with a small embedding budget**
[...]

PR #64, opened 2026-04-24
Search path: prefer-by-ID cache merge + sync-embed dedup (#62 B2)

tuky191:
## Summary
- Two structural fixes to stop sync-embedding the same candidate on every search frame.
- **Change 1**: prefer ID match over IoU when merging async embedding results. FTFTracker preserves IDs across frames; trust it and stop throwing away cached embeddings when box motion exceeds IoU=0.3.
- **Change 2**: short-retention sync-embed cache (\`recentSyncEmbeds\`). After sync fallback computes an embedding for candidate ID X, skip re-computing within 30 frames as long as the box hasn't drifted (IoU > 0.5).
- Stacked on #63 (A1). Merge #63 first.

## Measured impact (Snapdragon 870 / live camera, ~50s person-search session)

| | Total sync embeds | Unique IDs | Duplicate rate |
|---|---|---|---|
| Master baseline | ~150+ rejects/13s on same-person candidate | — | very high |
| This branch | **12 in 50s** | **12 (all unique)** | **zero** |

Every sync embed is now a genuinely new tracked ID. When an ID reappears with a similar box, the cached result is reused — no 23ms embed cost.

Process-thread search fps: 12-22fps (up from 9-18fps on comparable master scenes). Low-fps spikes are shorter and less frequent.

## Tracking quality
Same test session reacquired the locked person successfully **twice** (at 2 frames and 111 frames after LOST, sim=0.558). A pre-B2 comparable person session reacquired zero times in 13s and the user cleared in frustration.

## Why it's safe
- Cache entries are cleared on every state transition out of search (\`pendingEmbeddings = null\` path + \`clearLock\`).
- Stale entries are pruned each frame (>= SYNC_EMBED_CACHE_FRAMES old).
- IoU > 0.5 gate means the cached embedding is only reused when the crop is substantially the same — if the subject moves, we re-embed.
- Doesn't change what the ReacquisitionEngine sees — a reused cache hit is indistinguishable from a fresh sync embed, so scoring/gating behavior is identical. The cached sim reflects the crop that was embedded; if the subject has moved enough that the embedding wouldn't match anymore, the IoU gate prevents reuse.

## Test plan
- [x] Unit tests pass
- [x] Live-camera person search — 12 sync embeds in 50s, all unique IDs
- [x] Reacquires still fire (2× in test session)
- [ ] Video replay suite before merge

## What remains (not in scope here)
FTFTracker is assigning a new ID to essentially every re-detection during the test session — all 12 sync embeds were distinct IDs. That means the dedup cache's ID-based key doesn't help across re-IDs. Fixing tracker stability (IoU threshold, re-ID across occlusion, velocity prediction) is a separate follow-up.

### DIFF
--- a/app/src/main/java/com/haptictrack/tracking/ObjectTracker.kt
+++ b/app/src/main/java/com/haptictrack/tracking/ObjectTracker.kt
@@ -109,6 +109,23 @@ class ObjectTracker(
         val cachedBox: RectF? = null
     )
 
+    /**
+     * Sync-fallback dedup cache. During search the sync fallback was firing on the
+     * same candidate ID every frame (~12 rejects/sec = ~280ms/sec of sync embed),
+     * because async results had moved out of IoU>0.3 range or hadn't arrived in time.
+     * We remember the last sync result per ID and reuse it while the box is still
+     * close to where we embedded it. Cleared on every lock/reacq transition.
+     */
+    private data class SyncEmbedEntry(
+        val embedding: FloatArray,
+        val colorHistogram: FloatArray?,
+        val cachedBox: RectF,
+        val framesLost: Int
+    )
+    private val recentSyncEmbeds = mutableMapOf<Int, SyncEmbedEntry>()
+    private val SYNC_EMBED_CACHE_FRAMES = 30  // ~1-2s at typical search fps; IoU is the real staleness guard
+    private val SYNC_EMBED_CACHE_IOU = 0.5f  // box must overlap this much with cached box
+
     /** Callback: (displayObjects, lockedObject, imageWidth, imageHeight, contour) */
     var onDetectionResult: ((List<TrackedObject>, TrackedObject?, Int, Int, List<PointF>) -> Unit)? = null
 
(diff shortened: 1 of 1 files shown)


tuky191 (commented):
## Code Review: PR #64 — Search path: prefer-by-ID cache merge + sync-embed dedup (#62 B2)

229 tests pass. +83 −22, 1 file.

### No issues

Two clean fixes:

1. **Prefer-by-ID merge**: When merging async embedding results, match by detection ID first (trusting FTFTracker), fall back to IoU > 0.3 only for unmatched entries. This stops throwing away cached embeddings when box motion exceeds IoU threshold. The greedy IoU merge with `usedCacheKeys` is preserved as fallback.

2. **Sync embed dedup cache** (`recentSyncEmbeds`): Maps detection ID → `SyncEmbedEntry` with cached box and frame number. Reusable if IoU > 0.5 and within 30 frames. Pruned each frame. Cleared on state transitions. Eliminates the 150+ redundant sync embeds per search session observed on master.

Cache lifecycle is correct: entries cleared on lock clear and when not searching (`pendingEmbeddings = null` path).

Good to merge (after #63).

Issue #65, opened 2026-04-25
FTFTracker ID instability during person search — fresh ID per re-detection


tuky191:
## Summary

During the B2 verification work (#62), live-camera person-search sessions showed FTFTracker minting a fresh detection ID for nearly every re-detection of the same physical person. This breaks any per-ID caching/dedup downstream and makes identity work strictly harder than it needs to be.

## Concrete observation (session 20260424_213458_person)

- ~50 seconds of person search after LOST.
- 12 sync embeds fired, **all 12 distinct IDs**: 53, 54, 56, 57, 66, 67, 68, 69, 70, 71, 72, 76.
- Visually it's the same person filling the frame the whole time — they didn't leave or get occluded for any sustained period.
- The ID churn happens because `FrameToFrameTracker.minIou = 0.2` doesn't survive realistic between-frame motion when:
  - The person is large in the frame (small motion = big box delta in absolute terms)
  - Detection has occasional skip frames (skip + motion = bigger gap)
  - Box jitter from the detector itself adds noise on top

When IoU between frame N's detection and frame N-1's tracked box drops below 0.2 even once, FTFTracker assigns a fresh ID and the prior ID's history is gone.

## Impact

1. **Defeats the #62 B2 dedup cache** for fast-moving subjects — every re-detection is "new" from the cache's perspective.
2. **Defeats label-based reasoning** that assumes ID continuity (e.g. \"if this candidate has been person for 5 frames, treat its embedding as more trustworthy\").
3. **Forces the embedding tier to do all the heavy lifting** for identity, which is exactly the regime where the embedder is weakest (#51).

## Possible directions

- **Velocity-aware IoU**: predict next frame's box from `VelocityEstimator` and IoU against the prediction, not the last raw box.
- **Lower the IoU floor + add a centroid distance fallback**: keep an ID alive if the box is close enough by center even when IoU is low.
- **Multi-frame stitching**: keep \"recent dropped\" boxes alive for K frames and re-attach if a new detection IoU>X with the dropped one.
- **Detector skip awareness**: when the detector skipped frames (#62 A1 widens the skip window), expect bigger box deltas and relax the threshold for the first detection after a skip.

## Why this isn't in #62

#62 is detector-side perf and search-path optimization. Tracker stability is a separate signal-processing problem — fixing it independently lets us measure each lever in isolation.

## How to reproduce

[...]

PR #66, opened 2026-04-25
lockOnObject: move ML burst off processing thread (#62 D1)

tuky191:
## Summary
- Move the ~150-250ms lock-time ML burst (5 augmented embeddings + segmenter + person classifier + OSNet re-id + MobileFaceNet + scene negatives) off the processing thread.
- Snapshot bitmap + state under \`lastFrameLock\`, release immediately, run ML on a dedicated \`lockExecutor\` thread.
- Processing thread picks up the completed \`LockResult\` via \`AtomicReference\` and applies state mutations on the next frame.
- Stacked on #64 (B2). Merge #63 → #64 → this in order.

## Why
Pre-D1, \`lockOnObject\` held \`lastFrameLock\` for the full ~150-250ms ML burst, blocking the processing thread (which also acquires \`lastFrameLock\` to update \`lastFrameBitmap\`). Result: viewfinder froze on tap, and \`BitmapRing\` pool overflow allocations happened during the window because the processing thread couldn't return frames to the pool.

## What stays on the processing thread
- VT init (~30-50ms) — kept here to avoid touching \`VisualTracker\`'s thread-safety. If we want to move this off too, we'd need synchronization in \`VisualTracker.init/update/stop\`. Worth a follow-up if VT init shows up as a bottleneck.
- All state mutations (\`reacquisition.lock\`, debug capture, scenario recorder, scene-negative gallery additions). These are serialized through \`applyPendingLockIfAny()\` which runs on the processing thread.

## Bitmap ownership
- Snapshot bitmap is owned by the in-flight \`LockResult\`.
- A new lock cancels the prior future and recycles its (yet-unapplied) snapshot.
- \`clearLock\` cancels and recycles.
- \`shutdown\` shuts down the executor and recycles any pending bitmap.

## Test plan
- [x] Unit tests pass
- [x] Subjective: lock latency feels negligible — no viewfinder freeze on tap
- [x] Tracking quality preserved across multiple sessions:
  - Cup session: 7+ reacquires (sim 0.60-0.96)
  - Bowl session: bowl→cup label-flicker override fired (sim=0.96)
  - Chair session: 5 reacquires (16, 16, 17, 97, 1 frame latencies)
- [x] Gallery growth + scene negatives unchanged
- [ ] Edge case: clear during lock window — code path is correct (cancel + recycle), worth a manual repro

## What remains
- A3 (parallelize detector + embedder), B1 (sync embed batching), C1-C3 (GLES3 unlocks) — all in #62 tier 2/3.
- FTFTracker stability (#65) — orthogonal but compounds with everything in #62.

### DIFF
--- a/app/src/main/java/com/haptictrack/tracking/ObjectTracker.kt
+++ b/app/src/main/java/com/haptictrack/tracking/ObjectTracker.kt
@@ -126,6 +126,34 @@ class ObjectTracker(
     private val SYNC_EMBED_CACHE_FRAMES = 30  // ~1-2s at typical search fps; IoU is the real staleness guard
     private val SYNC_EMBED_CACHE_IOU = 0.5f  // box must overlap this much with cached box
 
+    /**
+     * Off-thread lock-on. The full lock burst (5 augmented embeddings + segmenter
+     * + person classifier + re-id + face + scene negatives) is ~150-250ms of GPU
+     * work that used to run on the processing thread under lastFrameLock, blocking
+     * frame processing for the full duration. Now we snapshot a bitmap, release
+     * the lock immediately, and run the ML on a dedicated thread; the processing
+     * thread picks the result up at the next frame boundary.
+     */
+    private val lockExecutor = java.util.concurrent.Executors.newSingleThreadExecutor { r -> Thread(r, "LockOnObject") }
+    private var pendingLockFuture: java.util.concurrent.Future<*>? = null
+    private val pendingLockResult = java.util.concurrent.atomic.AtomicReference<LockResult?>()
+
+    /** Result of off-thread lock ML; applied on the processing thread. */
+    private data class LockResult(
+        val trackingId: Int,
+        val boundingBox: RectF,
+        val label: String?,
+        val gallery: List<FloatArray>,
+        val colorHist: FloatArray?,
+        val personAttrs: PersonAttributes?,
+        val reIdEmb: FloatArray?,
+        val faceEmb: FloatArray?,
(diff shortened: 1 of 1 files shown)


tuky191 (commented):
## Code Review: PR #66 — lockOnObject: move ML burst off processing thread (#62 D1)

229 tests pass. +122 −39, 1 file.

### Low

**1. `applyPendingLockIfAny` runs VT init on the processing thread (~30-50ms)**

`visualTracker.init(result.snapshotBmp, result.boundingBox)` runs synchronously on the processing thread during `applyPendingLockIfAny`. This is documented (kept to avoid thread-safety issues with VisualTracker), but it's the remaining lock-time cost on the hot path. Worth noting for future optimization if VT init shows up in profiling.

- **Bitmap snapshot ownership is clean**: `snapshotBmp` is created under `lastFrameLock` (line 218), owned by `LockResult`, recycled in all exit paths: cancel (line 229), exception (line 264), successful apply (line 305, debug capture recycles).

- **Race between new lock and pending apply**: If the user taps lock again before the first `applyPendingLockIfAny` runs, `pendingLockFuture?.cancel(true)` kills the in-flight future, and `pendingLockResult.getAndSet(null)?.snapshotBmp?.recycle()` cleans up any completed-but-unapplied result. Correct.

- **Scene negatives use `snapshotDevRot` for coordinate remapping** (line 252-255): Correct — the snapshot bitmap is rotated, so detection boxes (screen space) need `mapToRotated` to crop correctly.

Good to merge (after #63 → #64).

PR #71, opened 2026-04-25
Two-people scenario fixture + identity research + camera-switch fix (#70)

tuky191:
## Summary
- Captures the failing-by-design two-people indoor scenario as a replay-test fixture (\`app/src/test/resources/scenarios/two_people_indoor.json\`).
- Adds three replay tests as regression baselines for #67 (Identity PR1: gate persons on OSNet).
- Includes \`docs/identity_research_2026-04-25.md\` — the full research report behind issues #51/#67/#68/#69.
- Bonus: fixes a real camera-switch crash uncovered while preparing the test session.

## The captured scenario

220 frames. Locked on a person near full-frame in an indoor scene where 16 frames have ≥2 persons visible. Live capture had 3 LOST/REACQUIRE cycles. The middle reacquire took **54 frames at sim=0.558 with reId=0.836** — i.e. OSNet correctly identified the same person while MobileNetV3 sat well below the gate. Exactly the failure mode #67 fixes.

## Tests added

- \`two people indoor - reacquires only as person\` — passes today (no wrong-category locks).
- \`two people indoor - reacquires multiple times\` — passes today (≥1 reacquire in replay).
- \`two people indoor - tracking rate baseline\` — passes today at 6%; comment explains expected post-#67 improvement.

These tests intentionally use loose thresholds because **scenario replay underestimates the live system** (async embedding artifacts, many candidates lack embeddings in the captured JSON). The point is to lock in current behavior so #67 has a regression baseline. Tighten thresholds after #67 lands.

## Camera-switch crash fix (independent)

\`SurfaceTextureFrameReader.stop()\` was recycling viewfinder bitmaps that Compose was still painting:

\`\`\`
java.lang.RuntimeException: Canvas: trying to use a recycled bitmap
  at BitmapPainter.onDraw
\`\`\`

Pre-existing issue from when the viewfinder pool was added (PR #60) — not introduced by recent perf work but uncovered when users actually exercised camera switch on the unified pipeline. Fix: drop inflight references on stop but don't recycle; GC handles them once Compose moves on.

## Test plan
- [x] Unit tests pass (3 new + all 184 existing)
- [x] Build + deploy to device
- [x] Camera switch front↔back↔front no longer crashes
- [x] Two-people scenario captured and replayable
- [ ] After #67 lands: re-run \`two people indoor - tracking rate baseline\`, expect material improvement, tighten threshold

## Stacking
[...]

Issue #70, opened 2026-04-25
Capture two-people and two-cars scenarios as replay test fixtures


tuky191:
**Parent**: #51.
**Blocks**: #67, #68, #69 (Identity PR1/PR2/PR3).

## Why

The identity research (\`docs/identity_research_2026-04-25.md\`) made specific claims about failure modes that we currently can't validate deterministically:

- Two-people scene: locked person A, person B walks through, lose lock, reacquire — needs to pick A not B.
- Two-cars scene: locked on red sedan A, pan to red sedan B nearby, lose lock, reacquire — needs to pick A not B.

We have no captured scenarios for these. The PR1/PR2/PR3 hypotheses can't be falsified without them — every \"this should work\" claim becomes \"trust me, it should work.\"

## What to capture

1. **Two-people-indoor** (e.g. living room, two adults visible at lock + during search):
   - Lock on person A.
   - Pan camera away briefly so person B is in frame at LOST time.
   - Pan back so both are visible.
   - Verify who gets reacquired.
   - Pull \`scenario.json\` + debug frames.

2. **Two-people-outdoor** (e.g. park, sidewalk):
   - Same flow with different lighting / scale conditions.

3. **Two-cars-outdoor** (e.g. parking lot, street):
   - Lock on car A (specific make/model/color).
   - Camera pans to car B (different car, possibly similar color).
   - Reacquire: should pick A not B.

4. **Two-cars-different-sizes** (e.g. sedan vs SUV in same scene):
   - Specifically tests size + color discrimination beyond color alone.

For each: produce \`scenario.json\` capture, copy to \`app/src/test/resources/scenarios/\`, write replay test asserting expected reacquire ID.

## How to capture

Standard scenario capture flow (CLAUDE.md \"Scenario Replay Testing\" section):
\`\`\`
adb install -r app/build/outputs/apk/debug/app-debug.apk
# trigger the scenario, clear lock when done
adb pull /sdcard/Android/data/com.haptictrack/files/debug_frames/session_<timestamp>/scenario.json
cp scenario.json app/src/test/resources/scenarios/<descriptive_name>.json
\`\`\`

Then write the test in \`ScenarioReplayTest.kt\` using \`assertReacquiresLabel(result, \"person\"|\"car\")\` plus an additional ID-level assertion that distinguishes A from B.

## Acceptance criteria

- 4 captured scenarios committed to \`app/src/test/resources/scenarios/\`.
[...]

Issue #69, opened 2026-04-25
Identity PR3: replace MobileNetV3-Large with MobileCLIP-S2 / DINOv2-small as generic embedder


tuky191:
**Source**: \`docs/identity_research_2026-04-25.md\` sections C.4, B.1, D (PR3).
**Parent**: #51.
**Depends on**: PR1 (#67), PR2 (#68).
**Confidence**: medium — depends on TFLite conversion success on Adreno 740.

## Hypothesis
MobileNetV3-Large is a 2019 ImageNet-classification model repurposed for similarity. It cannot meaningfully discriminate \"red car A\" from \"red car B\" or \"black mug A\" from \"black mug B.\" Self-supervised foundation models (DINOv2, MobileCLIP, SigLIP) trained explicitly for instance-level visual similarity are dramatically stronger on these cases. With OSNet covering persons (PR1) and adaptive floors covering cold-start (PR2), the remaining failure mode is rigid-object identity — squarely MobileCLIP's wheelhouse.

## Direct evidence

From the research:
- MobileNetV3-Large is trained for category discrimination (\"is this a car?\"), not instance discrimination (\"is this *the same* car?\").
- Two cars of different size + color get sim values that don't separate cleanly.
- Modern SoTA on-device options:
  - **MobileCLIP-S2** (Apple, CVPR 2024): 35.7M params, ~512-dim, designed for on-device. Competitive with CLIP-ViT-B/16 on retrieval. CoreML released; community ONNX→TFLite conversion path exists.
  - **DINOv2-small** (Meta, 2023): 21M params, 384-dim. Strong instance-retrieval performance without fine-tuning. Community PyTorch→ONNX→TFLite conversions exist.

## What to ship

1. Convert MobileCLIP-S2 image encoder to TFLite. Path: PyTorch checkpoint → ONNX → onnx2tf → TFLite, run via raw TFLite GPU (same pattern as \`magic_touch.tflite\` and OSNet, NOT via MediaPipe ImageEmbedder).
2. Ship \`mobileclip_s2_image.tflite\` (~50-80MB FP16) — this dominates the APK size impact.
3. Wire as drop-in replacement in \`AppearanceEmbedder\`. Segmenter integration and gallery-augmentation logic unchanged.
4. Re-tune the OSNet+MobileCLIP fusion in \`scoreCandidate\`: OSNet stays primary for persons, MobileCLIP becomes primary for non-persons.
[...]

Issue #68, opened 2026-04-25
Identity PR2: per-lock adaptive embedding floor + frozen-asset negative pool


tuky191:
**Source**: \`docs/identity_research_2026-04-25.md\` sections C.3, C.6, D (PR2).
**Parent**: #51.
**Depends on**: PR1 (file linked above).
**Confidence**: medium-high.

## Hypothesis
Even with OSNet gating for persons, fixed floors (\`MIN_EMBEDDING_SIMILARITY = 0.15\` and adaptive 0.3-0.5) are too coarse. A floor learned per-lock from the gallery's own intra-spread + a frozen scene-negative pool fixes the cold-start gap (no negatives, no learned boundary in the first ~30 search frames) and adapts to the actual embedder behavior on this lock.

## Direct evidence

From the research report:
- Adaptive floor uses \`_minGallerySim * 0.75\` clamped \`[0.3, 0.5]\`. \`_minGallerySim\` is computed only over augmented + accumulated gallery, which becomes very tight after VT runs (similar viewpoints). For person crops, augmented entries embed quite differently from each other, so on lock the floor clamps at 0.3 and stays.
- The \`OnlineClassifier\` already does small-MLP metric learning but requires runtime negatives, which the failure case can't generate (no positives ever pass the floor).
- Exemplar-SVM (Malisiewicz 2011) is the right structural fix: ship pre-computed background negative embeddings as an asset, train SVM at lock time, gate on SVM score.

## What to ship

1. Pre-compute ~1500 MobileNetV3 embeddings of generic indoor + outdoor background patches (random ImageNet/ADE20K crops). Ship as \`assets/frozen_negatives_mnv3.bin\`, ~7MB.
2. Same for OSNet, ~5MB → \`frozen_negatives_osnet.bin\`.
3. In \`ReacquisitionEngine.lock()\`: immediately train \`OnlineClassifier\` against \`gallery (positive) + frozen_negatives (negative)\`. Engine starts in classifier-active mode from frame 1 instead of waiting for ≥3 positives + ≥3 negatives.
4. Replace the \`MIN_EMBEDDING_SIMILARITY\` constant with: \`floor = max(0.15, mean(galleryToNegativeSims) + 0.5 * std(galleryToNegativeSims))\`, clamped \`[0.20, 0.55]\`. Same on the OSNet path with its own asset.
5. Tooling: a script in \`tools/\` to regenerate the frozen negatives if/when the embedder model changes.

## Cost estimate
[...]

PR #72, opened 2026-04-25
Identity: gate persons on OSNet, fix self-poisoning negatives (#67)

tuky191:
## Summary
- Gate person candidates on OSNet (a real re-ID model) instead of MobileNetV3 (a generic ImageNet classifier blind to person-instance identity).
- Stop adding rejected candidates to the negative pool during search — pre-fix, the locked person was being repeatedly added as a negative and the system was training itself to reject the right answer.
- Compute OSNet for every person candidate during search, not just top-2 by MobileNetV3. Without this, the wrong person could rank higher on MobileNetV3 and OSNet wouldn't run on the right one.
- Stacked on #71. Merge order: #63 → #64 → #66 → #71 → this.

## The bug

Live-camera capture (\`session_20260424_203704_person\`):
- Locked on a person, near full-frame.
- After LOST, FrameToFrameTracker correctly matched the same physical person (id=234) for 50+ frames at IoU 70-88%.
- ReacquisitionEngine rejected **all 97** of them at \`(hard threshold)\`, sim ∈ [0.08, 0.34] (MobileNetV3).
- OSNet \`reIdEmbedding\` was being computed every frame. Cosine to the lock-time OSNet embedding sat at 0.836 — strong same-person signal — and **was thrown away unread** because the candidate never passed the MobileNetV3 floor.

The whole identity stack downstream of MobileNetV3 was effectively dead code in the failure case.

## On-device video replay results (Snapdragon 870)

| Test | Pre-PR1 baseline | This PR |
|---|---|---|
| \`man_desk_camera_swing\` | 78% | 79–81% (within run-to-run noise) |
| \`person_playground\` | 90% | **99%** |
| \`boy_indoor_wife_swap\` | 30% threshold (known buggy: \"jumps from son to wife\") | **86%** |

The \`boy_indoor_wife_swap\` result is the headline. That test had a 30% tracking-rate threshold deliberately set low because of the documented \"tracking jumps from son to wife\" bug. PR1 brought it to 86% — nearly 3× the threshold, with 17 reacquires over 1347 frames all correctly labeled \`person\` and long stable stretches between losses.

## What changed (line-level)

**\`ReacquisitionEngine.scoreCandidate\` (lines ~500-555)**:
- Compute \`hasReId = lockedIsPerson && candidateIsPerson && both have reIdEmbedding\` early, before the gate.
- For \`hasReId\` candidates, set \`appearanceScore = cosineSimilarity(lockedReIdEmbedding, candidate.reIdEmbedding)\` (OSNet).
- Otherwise unchanged: \`appearanceScore = bestGallerySimilarity(candidate.embedding)\` (MobileNetV3).
- New floor \`PERSON_REID_FLOOR = 0.45f\` for the OSNet path. Tuned for Market-1501-style cosine distribution: same-person ≥ 0.5, different-person ≤ 0.4. Live-capture data point: same-person reId=0.836 vs MobileNetV3 sim=0.558.
- \`geometricOverride\` and \`labelOverride\` now use the effective \`appearanceScore\` (OSNet for person-person), so a strong re-ID match correctly bypasses position/size hard filters during fast camera movement.

**\`ReacquisitionEngine.scoreCandidate\` (rejection paths)**:
- Removed \`addNegativeExample(candidate.embedding!!)\` calls at the embedding-floor reject and the person/non-person mismatch reject. Negatives are still collected during confirmed VT tracking via \`addSceneNegative\` with its 0.85 sim filter — that path is correct.

**\`ObjectTracker.computeEmbeddingsSync\`**:
- Removed the \`.take(2)\` filter on person candidates. OSNet (and face when \`lockedFaceEmbedding != null\`) is now computed for every person candidate with an embedding.

**\`ObjectTracker\` sync embedding fallback**:
- For person candidates, also computes OSNet (and face) sync alongside MobileNetV3, so the new gate works end-to-end on freshly-detected persons. Sync result cache (\`SyncEmbedEntry\`) extended to hold the extra embeddings.

## Why this is safe

- \`hasReId\` requires both sides to have OSNet embeddings. If OSNet isn't computed (non-person, or async hasn't caught up and sync fallback didn't fire), the path falls back to the existing MobileNetV3 floor logic. No behavior change for non-person locks.
- The negative-poisoning fix is strictly less aggressive than before. We can no longer accidentally train against the locked person; we never train aggressively for separation either, which is fine because the gate now does the heavy lifting.
- 229 existing unit tests pass. One test was updated to reflect the new gate behavior (clearly-mismatching reId is now hard-rejected; previously was scored low in the weighted average — semantically equivalent intent).

## Test plan
- [x] Unit tests pass (229)
- [x] On-device video replay: man_desk, person_playground, boy_indoor_wife_swap — all pass with material improvement
- [x] Live tracking on device: lock on person, walk around, reacquire correctly
- [ ] Capture a fresh two-people indoor scene + add as scenario fixture → tighten \`two_people_indoor.json\` thresholds (left for #70 follow-up)

## What this does NOT fix (out of scope)
- Vehicles / two-cars confusion → that's #69 (PR3, embedder swap).
- Per-lock adaptive floor → that's #68 (PR2).
- FTFTracker ID instability → that's #65.

### DIFF
--- a/app/src/main/java/com/haptictrack/tracking/ObjectTracker.kt
+++ b/app/src/main/java/com/haptictrack/tracking/ObjectTracker.kt
@@ -119,6 +119,8 @@ class ObjectTracker(
     private data class SyncEmbedEntry(
         val embedding: FloatArray,
         val colorHistogram: FloatArray?,
+        val reIdEmbedding: FloatArray? = null,
+        val faceEmbedding: FloatArray? = null,
         val cachedBox: RectF,
         val framesLost: Int
     )
@@ -687,8 +689,8 @@ class ObjectTracker(
                         val reusable = cached != null &&
                             computeIou(best.boundingBox, cached.cachedBox) > SYNC_EMBED_CACHE_IOU
 
-                        val (embedding, hist) = if (reusable) {
-                            cached!!.embedding to cached.colorHistogram
+                        val syncEntry: SyncEmbedEntry? = if (reusable) {
+                            cached
                         } else {
                             // Compute embedding synchronously (raw crop fallback, ~8-23ms)
                             val embResult = appearanceEmbedder.embedAndCrop(bitmap, best.boundingBox, fallback = true)
@@ -697,22 +699,35 @@ class ObjectTracker(
                                 computeColorHistogram(embResult.maskedCrop, fullBox).also { embResult.maskedCrop.recycle() }
                             } else computeColorHistogram(bitmap, best.boundingBox)
                             if (embResult.embedding != null) {
-                                android.util.Log.d("EmbedSync", "Sync embedding for id=${best.id} label=${best.label}")
-                                recentSyncEmbeds[best.id] = SyncEmbedEntry(
+                                // Person candidates also need OSNet (and face when locked
+                                // has a face) so the new OSNet-gated path (#67) can fire
+                                // on freshly-embedded candidates without waiting for async.
+                                val isPersonCandidate = lockedIsPerson && best.label == "person"
+                                val reIdEmb = if (isPersonCandidate) personReId.embed(bitmap, best.boundingBox) else null
+                                val faceEmb = if (isPersonCandidate && reacquisition.lockedFaceEmbedding != null) {
+                                    faceEmbedder.embedFace(bitmap, best.boundingBox)
+                                } else null
+                                android.util.Log.d("EmbedSync", "Sync embedding for id=${best.id} label=${best.label} reId=${reIdEmb != null} face=${faceEmb != null}")
+                                val entry = SyncEmbedEntry(
                                     embedding = embResult.embedding,
                                     colorHistogram = computedHist,
+                                    reIdEmbedding = reIdEmb,
+                                    faceEmbedding = faceEmb,
                                     cachedBox = RectF(best.boundingBox),
                                     framesLost = currFramesLost
                                 )
-                            }
-                            embResult.embedding to computedHist
+                                recentSyncEmbeds[best.id] = entry
+                                entry
+                            } else null
                         }
 
-                        if (embedding != null) {
+                        if (syncEntry != null) {
                             merged.map { obj ->
                                 if (obj.id == best.id) obj.copy(
-                                    embedding = embedding,
-                                    colorHistogram = hist
+                                    embedding = syncEntry.embedding,
+                                    colorHistogram = syncEntry.colorHistogram,
+                                    reIdEmbedding = syncEntry.reIdEmbedding ?: obj.reIdEmbedding,
+                                    faceEmbedding = syncEntry.faceEmbedding ?: obj.faceEmbedding
                                 ) else obj
                             }
                         } else merged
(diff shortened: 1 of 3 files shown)


tuky191 (commented):
## Code Review: PR #72 — Identity: gate persons on OSNet, fix self-poisoning negatives (#67)

229 tests pass. +90 −47, 3 files. The headline PR in the stack.

### No high-severity issues

The design is sound: OSNet is a real re-ID model trained for person-instance discrimination, while MobileNetV3 is a generic ImageNet classifier blind to it. Gating person candidates on OSNet instead of MobileNetV3 is the structurally correct fix.

### Medium

**1. `computeEmbeddingsSync` now runs OSNet on ALL person candidates, not top-2** (`ObjectTracker.kt:970-990`)

Previously filtered to top-2 by MobileNetV3 similarity. Now runs OSNet + PersonAttributeClassifier + (optionally) FaceEmbedder on every person detection with an embedding. With `maxResults=5` from the detector, worst case is 5 person OSNet calls per frame on the async thread.

OSNet is ~8-12ms per call (GPU), so this adds ~24-36ms per async frame in the worst case (3 extra persons). The async pipeline runs one frame behind, so this doesn't directly impact processing-thread latency — but it extends the async computation time, which means results take longer to arrive and the sync fallback may fire more often during the gap.

This is the correct tradeoff (missing the right person from OSNet computation was the bug), but worth monitoring on scenes with many persons.

**2. Ratio test still uses MobileNetV3 `bestGallerySimilarity`, not OSNet** (`ReacquisitionEngine.kt:473-486`)

`findBestCandidate` applies the Lowe's ratio test using `bestGallerySimilarity(candidate.embedding!!)` — always MobileNetV3. For person candidates gated on OSNet, this means the ratio test operates on a different similarity signal than the one that drives scoring. Two persons with different OSNet scores but similar MobileNetV3 scores would be rejected as ambiguous, even if OSNet clearly distinguishes them.

Not a regression (ratio test was always MobileNetV3), but now that scoring and gating use OSNet, the ratio test is operating on a mismatched signal. Consider switching to the effective appearance score for the ratio test in a follow-up.

### Low

**3. Sync fallback `SyncEmbedEntry` extended but cache IoU check unchanged** (`ObjectTracker.kt:689`)

The `SyncEmbedEntry` now carries `reIdEmbedding` and `faceEmbedding`. The cache reuse gate (`IoU > 0.5`) checks box similarity, which is reasonable for MobileNetV3 (crop-sensitive), but OSNet is more robust to crop variation. A cached OSNet embedding could be valid even at IoU < 0.5. Not worth changing now — the 0.5 threshold is conservative-correct.

**4. `PERSON_REID_FLOOR = 0.45f` is a single data point** (`ReacquisitionEngine.kt:49-52`)

Tuned from one live capture (same-person 0.836 vs different-person gap). The Market-1501 distribution comment provides theoretical backing, but the floor sits right in the gap between 0.4 (different-person max) and 0.5 (same-person min). A tight margin — consider collecting more data points across varied lighting/clothing before tightening to 0.5 in #68.

- **Self-poisoning negative fix is critical**: The `addNegativeExample` calls were removed from both the embedding-floor reject (line 548) and the person/not-person mismatch (line 611). The failure mode was real: locked person → rejected by floor → added as negative → classifier/prototype trains against correct person → harder to reacquire next frame. Negatives still flow from `addSceneNegative` during confirmed VT tracking (0.85 sim filter). Correct.

- **`gateActive` replaces `hasAppearance` in override checks** (lines 558-559): Now OSNet can drive geometric/label overrides for person candidates. A strong OSNet match (>0.55) correctly bypasses position/size gates during fast camera movement. Consistent with the dual-path design.

- **Embedding requirement gate has OSNet exception** (line 530): `if (hasEmbeddings && candidate.embedding == null && !hasReId) return null`. If a person candidate has OSNet but no MobileNetV3 embedding, it passes the gate. This is correct — OSNet alone is sufficient for person identity.

- **Test update is semantically correct**: `reId tier scores matching body higher than mismatching` → `reId tier rejects clearly-mismatching person via OSNet gate`. Same test vectors, but the assertion changed from "matchScore > wrongScore" to "matchScore != null, wrongScore == null". Reflects the gate behavior.

- **boy_indoor_wife_swap: 30% → 86%** is a dramatic improvement. 17 reacquisitions over 1347 frames, all correctly labeled `person`. This was the documented "jumps from son to wife" bug — OSNet correctly discriminates the two persons.

### Weight verification (all 4 scoring tiers)

Verified all weights sum to 1.0:

| Tier | Primary | Weights |
|------|---------|---------|
| Face | faceScore | 0.45 + 0.20 + 0.10 + 0.05·posConf + 0.10 + 0.05 + unused → 1.0 ✓ |
| Re-ID | reIdScore | 0.40 + 0.20 + 0.10·posConf + 0.05 + 0.15 + 0.10 + unused → 1.0 ✓ |
| Generic | embScore | 0.30/0.50 + 0.25/0 + 0.20/0 + 0.10·posConf + 0.10 + 0.10 + 0.05 + unused → 1.0 ✓ |
| Fallback | positionScore | 0.50·posConf + 0.25 + redistributed → 1.0 ✓ |

Good to merge (after #63 → #64 → #66 → #71).

tuky191 (commented):
## Follow-up review (commit f921bb3)

**#2 (Med) Ratio test uses MobileNetV3 not OSNet** — **Fixed.** New `effectiveAppearanceSim()` helper mirrors the OSNet-vs-MobileNetV3 split from `scoreCandidate`: person-person uses OSNet cosine, everything else uses MobileNetV3 gallery similarity. Both the ratio test and the gate/scoring now operate on the same signal. Comment updated to document this.

Good to merge.

Issue #51, opened 2026-04-22
Better embedding model and supplementary identity signals


tuky191:
## Context

Spun off from #50 (object identity). The core identity system is solid — tentative confirmation, ratio test, online classifier, gallery-relative threshold all working. Zero wrong reacquisitions across all test scenarios.

Remaining items that could further improve identity discrimination but weren't needed for the current quality bar.

## 1. DINOv2-small embedder

Replace MobileNetV3 (10MB, 1280-dim) with DINOv2-small (88MB, 384-dim). Benchmarked in #50 research:

| Model | Same-obj min sim | Cross-obj gap | Reacq accuracy |
|---|---|---|---|
| MobileNetV3 (current) | 0.318 | -0.067 FAIL | 2/3 |
| DINOv2-small | 0.626 | +0.024 POOR | 2/3 |
| DINOv2-base | 0.734 | +0.059 PASS | 3/3 |

TFLite conversion verified working via `litert-torch`. ~50-100ms inference on Adreno 740 GPU (vs 15ms for MobileNetV3). Would need async pipeline adjustment.

Tradeoff: +34-78MB APK, 3-6x slower embedding. May not be worth it given the current identity system handles discrimination well.

## 2. Template matching (pixel-level NCC)

Store actual pixel patches of the tracked object. Normalized cross-correlation between candidate crops and stored templates.

- Complementary to embedding similarity (pixel vs semantic)
- Fast on GPU, no ML needed
- Works well for rigid objects (cup, mouse), less for deformable (person)
- View-dependent — doesn't handle rotation well

## 3. Multi-modal fusion learning

Learn which identity signals matter for each specific object:
- Red ball → color histogram is highly discriminative
- Person → re-ID + face dominates
- Mouse on desk → shape/size matters most

Currently using fixed weights. Could train a small per-object weight vector from accumulated scoring data.

## 4. Contrastive learning at runtime

Fine-tune a lightweight MLP head on top of frozen MobileNetV3 features using triplet loss (anchor=locked, positive=same object, negative=scene). Adapts embeddings to the specific object.

Heavy: requires gradient computation on-device. Speculative.

## 5. SAM2 for segmentation

Replace magic_touch with SAM2 for much better instance masks. Better masks → better embeddings (less background contamination).

Separate effort — would need ONNX/TFLite conversion and GPU benchmarking.

## Priority

Items 1-2 are most actionable. Items 3-5 are speculative and should only be pursued if 1-2 don't close remaining gaps.

[...]

PR #79, opened 2026-04-25
Identity: per-lock adaptive embedding floor + classifier cold-start (#68)

### LINKED ISSUE #77: PERSON_REID_FLOOR validation across more captures
**Source**: PR #72 review (Low #4).

## Context
\`PERSON_REID_FLOOR = 0.45f\` was tuned from a single live capture (\`session_20260424_203704_person\`) where same-person OSNet hit 0.836 and MobileNetV3 sat at 0.558. Market-1501 literature provides theoretical backing for the typical same-person ≥ 0.5 / different-person ≤ 0.4 distribution, but 0.45 sits exactly in the gap.

A more conservative floor (0.5+) might miss legit reacquires on harder cases (motion blur, low light, partial occlusion). A more permissive floor (0.4) might leak wrong-person matches.

## Why deferred
PR #72 ships with a theoretically reasonable floor backed by one captured data point. The on-device video replay results (boy_indoor_wife_swap 30%→86%) validate that the floor isn't too high. But edge cases haven't been tested.

## What to do
1. Capture 5-10 more two-people scenarios across:
   - Lighting (bright indoor / low indoor / outdoor / mixed)
   - Subject distance (close / medium / far)
   - Pose variation (turning / walking away / partial occlusion)
2. For each, log OSNet sim values during search frames where the engine considered candidates.
3. Build a histogram of same-person vs different-person OSNet sims across the dataset.
4. Either confirm 0.45 sits in the gap, or adjust.

## Verification target
On the captured dataset, the floor should produce:
- ≥ 90% recall for same-person at typical scoring (lock vs same person across angles)
- ≤ 5% false-pass for different-person where the wrong person is also in frame

## Related
This is a precondition for confidence in #68 (Identity PR2: per-lock adaptive floor). PR2 may obviate this issue if the per-lock floor learns the gap from gallery + frozen negatives.

tuky191:
## Summary
- Replace the static \`MIN_EMBEDDING_SIMILARITY = 0.15\` and \`PERSON_REID_FLOOR = 0.45\` thresholds with a per-lock floor learned from the gallery → frozen-negatives similarity distribution at lock time.
- Cold-start the \`OnlineClassifier\` at \`lock()\` using \`gallery (positive) + frozen (negative)\` so the classifier has a decision boundary from frame 1, not from frame 30+ of confirmed VT tracking.
- **Structural side only.** No asset bundled yet — the loader returns empty when the asset is missing and the engine falls back to existing static floors. Asset generation will be a follow-up commit on this branch.

## What "frozen negatives" are

Pre-computed embeddings of generic background-scene crops, bundled as a binary asset. ~1500 entries × 1280-dim (MobileNetV3) ≈ 7.7MB; ~1500 × 512-dim (OSNet) ≈ 3MB. They never change after build — they're a fixed reference distribution of "what arbitrary stuff in the world looks like through this embedder."

At lock time we derive the floor as \`mean + 0.5σ\` of the gallery → frozen sim distribution. This is "the false-match similarity expected for random stuff" + a small safety margin. Anything above this is meaningfully more similar than scene noise.

## Why this matters

Static floors are tuned for "average" locks but wrong for any specific lock:
- Red mug on a white table → needs a tight floor (~0.3+; anything below is clearly not the mug).
- Person in a busy scene → needs a permissive floor (~0.45-0.5; same person can dip to 0.3-0.4 on hard frames).
- Today both get 0.40 (MobileNetV3) / 0.45 (OSNet).

The adaptive floor reads the actual lock and produces an appropriate threshold for it.

## Changes

**\`FrozenNegatives.kt\`** (new, 70 lines): binary loader. Format documented in the file header. Returns empty list on missing/malformed asset (graceful fallback).

**\`ReacquisitionEngine.kt\`** (+90 lines):
- Two new optional constructor params for the negative pools (default empty).
- \`companion object.create(context)\` factory that loads both assets via \`FrozenNegatives\`. ObjectTracker now uses this factory.
- At \`lock()\`: computes adaptive floors (clamped \`[0.20, 0.55]\` MobileNetV3 / \`[0.30, 0.60]\` OSNet), cold-starts the classifier.
- At \`scoreCandidate\`: \`_adaptiveOsnetFloor ?: PERSON_REID_FLOOR\` / \`_adaptiveMobileNetFloor ?: <existing gallery-relative logic>\`. Existing behavior preserved when asset missing.
- At \`clear()\`: resets both floors.

**\`ObjectTracker.kt\`** (+1 -1): \`val reacquisition = ReacquisitionEngine.create(context)\` instead of bare default.

**Tests**: 3 new (232 total, all pass).

## Test plan
- [x] Unit tests pass (232)
- [x] Production build succeeds (loader handles missing asset gracefully)
- [ ] Asset generation follow-up: \`tools/generate_frozen_negatives.py\`, source ADE20K val (~2000 indoor+outdoor scene crops), embed via MediaPipe + OSNet TFLite, save 1500 diverse rows to \`app/src/main/assets/frozen_negatives_{mnv3,osnet}.bin\`.
- [ ] Once asset lands: re-run video replay suite, expect:
  - \`boy_indoor_wife_swap\` stays at 86%+ (no regression from PR1)
  - \`person_playground\` stays at 99%+
  - Cup/chair scenarios stay or improve (per-lock floor better than the static 0.40)

## Stacking
Standalone PR off master. The 5-PR identity stack we landed earlier (#63 → #64/#78 → #66 → #71 → #72) is fully merged.

## Related issues
- Closes part of #51 (Better embedding model and supplementary identity signals — adaptive floor is the supplementary mechanism).
- Resolves #77 (PERSON_REID_FLOOR validation) — adaptive floor obviates the need to hand-tune a single global value.
- #68 itself.


tuky191:
## On-device verification (2026-04-25)

Built + deployed on Snapdragon 870. Both assets loaded successfully and adaptive floors activated as designed.

### Asset load
\`\`\`
FrozenNeg: frozen_negatives_mnv3.bin loaded: 1500 × 1280
FrozenNeg: frozen_negatives_osnet.bin loaded: 476 × 512
\`\`\`

### Bowl lock (non-person, MNV3 path)
\`\`\`
Reacq: LOCK id=6 label="bowl" ... gallery=5 mnv3Floor=0.200
\`\`\`
Floor clamped at the bottom of [0.20, 0.55] — gallery is well-distinct from generic scene noise. Pre-PR2 this would have been the static \`MIN_EMBEDDING_SIMILARITY = 0.15f\`-clamped-to-0.40 fallback; the adaptive path picked a slightly tighter 0.20 floor for this lock.

### Person lock (OSNet path)
\`\`\`
Reacq: LOCK id=131 label="person" ... attrs="man, short sleeves, shorts/skirt, orange/blue" mnv3Floor=0.200 osnetFloor=0.600
\`\`\`
**OSNet floor at 0.600** — clamped at the **top** of the [0.30, 0.60] band. The formula \`mean + 0.5σ\` of \`lockedReIdEmbedding → frozen_osnet\` similarities exceeded 0.60 and got pulled back. Tighter than the static \`PERSON_REID_FLOOR = 0.45f\` it replaced.

In this test session all five reacquires of the same person landed at reId 0.778 / 0.885 / 0.864 / 0.890 / 0.865 — well above 0.600, so no false rejections. But this is worth flagging as a known limitation:

**Known limitation**: the OSNet negative pool was generated from past session captures of the same recurring test subject (me/the device user). Despite the 0.98 diversity dedup at generation time, multiple angles of the same person get past it. That biases the floor up — the gallery looks more distinct from the negatives than it should, because some "negatives" are actually the same person at different angles.

Mitigations / follow-ups:
1. Re-generate the OSNet negative pool from a more diverse source (e.g. add captures from different humans, downsample by face similarity).
2. Lower the OSNet upper clamp from 0.60 to 0.55 if real-world reacquires start failing in the 0.45-0.60 reId band.
3. Both — they're complementary.

For now, shipping as-is: in the test session the floor was strict but didn't reject correctly. As more diverse captures land in subsequent generations of the negative pool, the formula will produce more accurate floors automatically.

### Tracking quality preserved
Five reacquires in ~10s of camera panning, all label=person, none reaching timeout. No regression visible.

### Drift detection slow on a separate ~7s case
While testing this PR I observed an unrelated drift-detection delay — filed as #80. It's not a regression from PR2; the same hysteresis-reset behavior existed before. Captured trace data is in #81.

tuky191 (commented):
## Code Review: PR #79 — Identity: per-lock adaptive embedding floor + classifier cold-start (#68)

232 tests pass (+3 new). +569 −9, 7 files. The biggest identity infrastructure PR since #52.

### Medium

**1. `computeAdaptiveFloor` runs 1500 × 5 × 1280 = 9.6M float ops at lock time** (`ReacquisitionEngine.kt:808-820`)

For MNV3: iterates 1500 frozen negatives, each calling `bestGallerySimilarity` against the 5-entry lock gallery. ~10ms on Snapdragon 870. This runs on the lock executor (off-thread, PR #66), so it doesn't block tracking. But it extends the lock burst from ~200ms to ~210ms.

Not a problem today, but if frozen negative count grows (the PR description mentions 1500 as a starting point), this scales linearly. Consider pre-computing the gallery→frozen matrix as a single batch dot product instead of N individual `bestGallerySimilarity` calls.

**2. Frozen negative assets add ~8.3MB to APK** (`frozen_negatives_mnv3.bin` 7.3MB + `frozen_negatives_osnet.bin` 952KB)

The MNV3 asset alone is 7.3MB (1500 × 1280 × 4 bytes). This is comparable to a TFLite model. For reference, the MobileNetV3 embedder model itself is 10MB. Consider:
- Quantizing to FP16 (halves size, negligible quality impact for normalized vectors)
- Reducing count if diminishing returns are found (e.g., 500 might produce similar `mean + 0.5σ`)
- Shipping only OSNet frozen negatives (952KB) if the MNV3 adaptive floor doesn't provide enough lift over the existing gallery-relative formula

**3. Classifier cold-start trains on 1500 frozen negatives** (`ReacquisitionEngine.kt:288-293`)

`OnlineClassifier.train()` runs 80 iterations of gradient descent over `gallery (5) + frozen (1500) = 1505` examples at 1280 dims. Each iteration: 1505 forward + backward passes × 1280-dim dot products. This is ~154M float ops per lock, running on the lock executor.

The L2 regularization and learning rate were tuned for the original regime (~12 positives + ~10 negatives = ~22 examples). With 1505 examples the gradient is dominated by negatives 300:1. The `gradB / n` and `gradW / n` normalization helps, but the massive class imbalance may undertrain on the positive class. Consider:
- Weighting positive examples by `negCount / posCount` in the gradient
- Or subsampling frozen negatives to ~50-100 for training (using the full set for floor computation only)

### Low

**4. OSNet frozen negatives only 476 entries vs 1500 MNV3** (`frozen_negatives_osnet.bin`)

The OSNet asset has 476 entries — likely limited by the number of person crops in the source scenarios. The adaptive floor with 476 samples is noisier than 1500. Not blocking — the [0.30, 0.60] clamp band provides guardrails.

**5. `_adaptiveOsnetFloor` uses single lock-time re-ID embedding as the "gallery"** (`ReacquisitionEngine.kt:279-282`)

`computeAdaptiveFloor(listOf(lockedReIdEmbedding!!), frozenNegativesOsnet)` computes the floor from a 1-entry "gallery". The `bestGallerySimilarity` degenerates to a single cosine similarity per negative. This is statistically fine (the mean+0.5σ computation is over 476 similarities), but the floor is entirely determined by one embedding's relationship to the frozen set — no gallery diversity to average out lock-time crop noise.

- **Binary format is correct**: Verified both assets — headers match (count, dim), total sizes match (8 + count × dim × 4). L2-normalization responsibility is on the Python generator, which is documented.

- **Fallback strategy is clean**: `_adaptiveMobileNetFloor ?: <existing gallery-relative logic>` preserves all existing behavior when assets are missing. Production path uses `ReacquisitionEngine.create(context)`, tests use the default constructor (no frozen negatives). No test regressions.

- **Floor clamp bands are conservative**: MNV3 [0.20, 0.55], OSNet [0.30, 0.60]. Even a badly-tuned adaptive floor can't go below 0.20 (MNV3) or 0.30 (OSNet), preventing accidental accept-everything. Upper bound prevents accidental reject-everything.

- **FrozenNegatives.kt loader is robust**: Handles missing asset (IOException → empty list), malformed header (size mismatch → empty list), wrong dimension (dim != expectedDim → empty list). All error paths logged.

- **`computeAdaptiveFloor` handles edge cases**: Returns `Float.NaN` on empty inputs, but callers always check `frozenNegatives.isNotEmpty()` before calling, and the result goes through `coerceIn()` which clamps NaN to the lower bound. Safe.

- **Generator script**: Two-mode design (OSNet from person scenario crops, MNV3 from random scene crops) with diversity filtering (cosine > 0.95 → drop). Well-documented. Lives in `tools/` with the other research scripts.

Good to merge. The class imbalance in classifier cold-start (#3) is the most impactful concern — worth a quick experiment with positive weighting or negative subsampling before the asset generation follow-up.

tuky191:
Addressed Medium #3 (class imbalance). Subsampling frozen negatives to 100 for classifier cold-start, full pool still used for adaptive floor statistics. Deterministic shuffle seeded by trackingId for test reproducibility. 232 tests pass.

Deferred:
- Medium #1 (batch dot product) — current cost is ~10ms one-time off-thread; revisit if frozen pool grows
- Medium #2 (FP16 quantization) — separate change since it requires regenerating both .bin assets via the Python tool. Tracked as follow-up issue if APK size becomes a real concern (currently +8.3MB on a ~120MB total APK, ~6%)

tuky191 (commented):
## Follow-up review (commit 1148212)

**#3 (Med) Classifier cold-start class imbalance** — **Fixed.** Subsamples frozen negatives to `CLASSIFIER_COLD_START_NEGATIVES = 100` for classifier training (5:100 = 1:20 ratio, down from 1:300). Full 1500 set still used for adaptive floor statistics. Deterministic shuffle seeded by `trackingId` for test reproducibility. Comment documents the rationale and references the review finding.

Good to merge.

Issue #80, opened 2026-04-25
Drift detection too slow when subject leaves view (~7s to declare LOST)


tuky191:
## Symptom

When a locked subject moves out of view (camera pan, subject walks off-screen), it takes ~7 seconds before the system transitions to LOST and starts searching. The visual tracker continues reporting high confidence on a phantom box; detector cross-check eventually fires after \`VT_MAX_UNCONFIRMED = 10\` frames.

User-visible effect: prolonged stale tracking. The bounding box stays drawn in the wrong place, haptics keep firing on a non-existent subject, auto-zoom holds on empty space.

## Concrete trace

From session 2026-04-25 11:50:

\`\`\`
11:50:39.692  VisualTracker INIT box=[0.07,0.41,0.98,1.00]
11:50:42.442  VisualTracker: Template mismatch 1/3 (sim=0.341)   ← 2.7s after lock
11:50:43.678  VisualTracker: Template mismatch 1/3 (sim=0.398)   ← counter RESET between
11:50:46.801  VisualTracker: DRIFT detected — 11 unconfirmed frames, stopping  ← 7.1s
\`\`\`

Note both template mismatches log as **1/3** — the counter reset between them. The primary drift signal (template self-verification, 3 mismatches in a row) never tripped; the slower detector cross-check is what fired.

## Root cause

\`VisualTracker.kt\` template-mismatch counter has asymmetric hysteresis: increment when sim < \`TEMPLATE_SIM_LOW\` (0.4), reset to 0 when sim > \`TEMPLATE_SIM_OK\` (0.5). Hysteresis is correct in principle (avoids oscillation around a single threshold) but currently too aggressive — a single high-sim flicker between mismatches sends the counter to 0. When the subject is partially out of frame, the VT crop occasionally lands on a stable backdrop patch, resets the counter, and the 3-in-a-row threshold rarely satisfies.

So in practice the primary signal is dead, and we always fall back to detector cross-check at \`VT_MAX_UNCONFIRMED = 10\`. With detector running ~every 2nd-3rd processing frame, that's 5-7 seconds wall-clock.

## Why this isn't fixed by a threshold tweak

Per the project memory \`feedback_no_parameter_tweaking\`: lowering \`VT_MAX_UNCONFIRMED\` would speed up THIS case at the cost of false positives in genuinely-noisy scenes (small motion, brief occlusion). The structural problem is that the hysteresis-with-reset model treats brief flicker as conclusive evidence of "still locked", which it isn't.

[...]

PR #81, opened 2026-04-25
docs: drift-detection trace data for #80

tuky191:
## Summary
Two captured traces for #80 — slow (~7s) and fast (~3s) drift-detection cases on the same code, back-to-back. Frames + session.log + a README that walks through the contrast.

## What's here
- \`docs/issues/80-drift-detection/curated/slow_7s/\` — hysteresis-reset case, 7s lock-to-drift
- \`docs/issues/80-drift-detection/curated/fast_3s/\` — clean template 3-in-a-row case, 3s lock-to-drift
- \`docs/issues/80-drift-detection/curated/README.md\` — diagnosis + implications for the structural fix

Each curated dir has 3 files: \`session.log\` (timing), \`lock.png\` (locked subject), \`lost_at_drift.png\` (scene when DRIFT fired). 2.5MB total.

## What's NOT here
The full ~95MB raw pulls (every annotated frame from both sessions) live on disk only — gitignored under \`docs/issues/*/slow_7s_case/\` and \`docs/issues/*/fast_3s_case/\`. The curated subset is enough to diagnose the bug; the full set is available locally if anyone needs it.

## Why this PR
#80 (drift detection slow when subject leaves view) is filed but unfixed. Whoever picks it up needs the trace data to compare against — what the slow case looks like vs the fast case. Landing this lets the eventual fix be reviewed against concrete evidence rather than re-running the capture from scratch.

This is doc-only — no code changes besides \`.gitignore\` adding the raw-pull patterns.


tuky191 (commented):
## Code Review: PR #81 — docs: drift-detection trace data for #80

Doc-only PR. +349, 8 files, ~2.5MB of curated trace data.

### No issues

Clean diagnostic data for #80. Two curated traces (slow 7s vs fast 3s) with lock.png, lost_at_drift.png, and session.log each. README walks through the contrast. Raw ~95MB pulls correctly gitignored. Useful reference for validating #82's fix and any future drift detection changes.

Good to merge.

Issue #83, opened 2026-04-25
OSNet gate too lenient: wrong-person reacquire at reId 0.55-0.70 despite PR1


tuky191:
## Symptom

Locked on a person; the engine reacquired what the user identifies as a **different** person, with OSNet \`reId\` similarity in the 0.55-0.70 range — well above the static \`PERSON_REID_FLOOR = 0.45\` set by #67 (PR1).

## Concrete trace

Session 2026-04-25 13:00:10 (frames pulled to \`/tmp/wrong_person_session/session_20260425_130010_person/\`):

\`\`\`
13:00:10.309  LOCK id=4 label="person" attrs="man, short sleeves, shorts/skirt, orange/blue"
13:00:14.399  REACQUIRE id=9  sim=0.809 reId=0.735                 ← borderline
13:00:18.999  REACQUIRE id=14 sim=0.564 reId=0.659                 ← user says: wrong person
13:00:20.523  REACQUIRE id=15 sim=0.572 reId=0.660                 ← user says: wrong person
13:00:26.724  REACQUIRE id=24 sim=0.847 reId=0.854 face=0.753       ← correct person (face high)
13:00:36.567  REACQUIRE id=31 sim=0.854 reId=0.695 face=0.482       ← borderline
\`\`\`

Two of the five reacquires (id=14, id=15) had \`reId ~0.66\` and **no face captured** for the candidate. With no face to over-rule OSNet, the OSNet sim alone decided the gate. 0.66 > 0.45 = pass.

## Why PR1 isn't enough

PR1's structural fix was correct: gate on OSNet (a real re-ID model) instead of MobileNetV3 (a generic ImageNet classifier). But the static floor `PERSON_REID_FLOOR = 0.45f` was tuned from one capture where same-person hit 0.836 vs MobileNetV3 0.558 — implying a wide gap between same and different person. In practice on Snapdragon 870 / Adreno 740, OSNet produces:

- Same person across angles: 0.7-0.95 (confirmed by id=24, id=31 in this trace)
- **Different person at similar pose/clothing: 0.55-0.70** (this issue)

The 0.45 gate sits well below the different-person band, letting wrong-person matches through.

## Why PR2 alone isn't enough either

[...]

Issue #67, opened 2026-04-25
Identity PR1: gate persons on OSNet, fix negative-poisoning, compute OSNet for all person candidates


tuky191:
**Source**: \`docs/identity_research_2026-04-25.md\` section D (PR1) and section A.3.
**Parent**: #51.
**Confidence**: high — this is a coding gap fix, not a research bet.

## Hypothesis
Two-people failures are caused by gating on a generic ImageNet embedder (MobileNetV3-Large). If we gate on OSNet — which we already compute every search frame — for person-locked sessions, two-people reacquisition becomes reliable.

## Direct evidence the gap exists

Session \`session_20260424_203704_person\` (logs in \`/tmp/sessions_62_baseline/\`):
- FrameToFrameTracker correctly matched the same physical person (id=234) across 50+ frames at IoU 70-88%.
- ReacquisitionEngine rejected **all 97** of them with \`(hard threshold)\` at sim 0.08-0.34 (MobileNetV3).
- OSNet was computed for the candidate every frame and **thrown away unread** because the gate happened first.
- Cup session at the same time: sim 0.70-0.91, reacquired in 1-11 frames.

So the embedder works fine on rigid objects with stable appearance and falls off a cliff on people, exactly because MobileNetV3-Large is a category-discrimination model not an identity model.

## What to ship

1. In \`ReacquisitionEngine.scoreCandidate\` (lines 494-695): when \`lockedIsPerson && candidate.label == \"person\" && candidate.reIdEmbedding != null && reacquisition.lockedReIdEmbedding != null\`, derive \`appearanceScore\` from \`cosineSimilarity(lockedReIdEmbedding, candidate.reIdEmbedding)\` instead of \`bestGallerySimilarity(candidate.embedding)\`. Fall back to MobileNetV3 if OSNet isn't available for whatever reason.

2. In \`ObjectTracker.computeEmbeddingsSync\` (lines 959-977): compute OSNet for **all** person candidates during search, not just top-2 by MobileNetV3 sim. Without this, fix (1) doesn't help when MobileNetV3 ranks the wrong candidate higher.

3. In \`ReacquisitionEngine.scoreCandidate\` (lines 526, 582): **stop** calling \`addNegativeExample\` on rejected candidates during search. Move scene-negative addition into VT-confirmed tracking only (already partially there via \`addSceneNegative\`'s 0.85 filter). Today's behavior actively poisons the negative pool with the locked person.

[...]

PR #84, opened 2026-04-25
Identity: face embedding as a gate (#83 phase 1)

tuky191:
## Summary
PR1 (#67) gated person candidates on OSNet at \`PERSON_REID_FLOOR = 0.45\`. On-device empirical: OSNet produces same-person sim 0.7-0.95 and **different-person sim 0.55-0.70** — wrong-person matches in the 0.55-0.70 band leak through the 0.45 floor.

This PR adds face embedding as a hard gate. When both lock and candidate have face embeddings (MobileFaceNet, ArcFace-style), face VETOES OSNet — a face sim < FACE_FLOOR (0.40) hard-rejects regardless of reId match.

## On-device verification (Snapdragon 870, 2026-04-25)

After locking on a person, panned camera to a different person whose face was visible. The face gate fired 20+ times in 1.5s:

\`\`\`
FACE_GATE_REJECT: faceSim=0.089 < 0.400 (reIdSim=0.679)
FACE_GATE_REJECT: faceSim=0.260 < 0.400 (reIdSim=0.695)
FACE_GATE_REJECT: faceSim=0.064 < 0.400 (reIdSim=0.630)
FACE_GATE_REJECT: faceSim=0.022 < 0.400 (reIdSim=0.639)
\`\`\`

All reId values in [0.63, 0.70] (would have passed PR1's gate alone, the documented #83 failure case). All face values in [0.06, 0.26] (clearly different face). Veto behaved correctly.

## What changed
- New \`FACE_FLOOR = 0.40f\` constant.
- New face gate in \`scoreCandidate\` after the embedding floor: when \`lockedFaceEmbedding != null && candidate.faceEmbedding != null\` for person-person, compute face cosine sim. Below 0.40 → \`return null\` with a \`FACE_GATE_REJECT\` log.

## What this PR does NOT fix
The no-face case. If the candidate has no face data (BlazeFace failed, distance, occlusion), the face gate doesn't fire and OSNet sim alone decides. Same test session showed a reacquire at \`reId=0.639\` with no face data, which still passed.

Phase 2 (color hist + person attrs as ANDed gates) is the follow-up for that case. Tracked in #83.

## Why face is the right primary gate

MobileFaceNet was trained explicitly for face identity (ArcFace metric learning on millions of identities), with cleaner same/different-person separation than OSNet's whole-body re-ID. When face is visible, it's the strongest single human-identity signal we compute. The 0.40 threshold sits in the gap between the typical same-person band (≥ 0.5) and different-person band (≤ 0.3).

## Tests
229 unit tests pass. Two pre-existing tests updated:
- \`face tier scores matching face higher than mismatching\` → \`face tier rejects clearly-mismatching face via face gate (#83)\` (assertion changed from "matchScore > wrongScore" to "matchScore != null, wrongScore == null").
- \`face tier takes precedence over reId tier\` → \`face gate vetoes good body when face mismatches (#83)\` (same shift — face is now a gate, not a scoring weight).

## Test plan
- [x] Unit tests pass (229)
- [x] Build + deploy
- [x] On-device repro: pan camera to different person with face visible — face gate vetoes correctly
- [ ] Capture this scenario as a replay test fixture for regression
- [ ] Phase 2: color + attrs ANDed gates for the no-face case

### DIFF
--- a/app/src/main/java/com/haptictrack/tracking/ReacquisitionEngine.kt
+++ b/app/src/main/java/com/haptictrack/tracking/ReacquisitionEngine.kt
@@ -49,6 +49,14 @@ class ReacquisitionEngine(
          *  different-person sim in [0.2, 0.4]. 0.45 sits in the gap. Tuned from a live
          *  capture where same-person reId hit 0.836 while MobileNetV3 sim was 0.558. */
         const val PERSON_REID_FLOOR = 0.45f
+        /** Face embedding (MobileFaceNet, ArcFace-style) cosine floor for person identity.
+         *  Same-person face sim typically >= 0.5; different-person face sim typically
+         *  <= 0.3. 0.4 sits in the gap. When both lock and candidate have face
+         *  embeddings, face VETOES OSNet — even a high reId sim gets rejected if face
+         *  says different. Catches the case where OSNet's whole-body sim sits in
+         *  the wrong-person 0.55-0.70 band but face cleanly separates identity.
+         *  Filed as #83. */
+        const val FACE_FLOOR = 0.40f
         /** Maximum embeddings to keep in gallery. */
         const val MAX_GALLERY_SIZE = 12
         /** Maximum negative examples to store. */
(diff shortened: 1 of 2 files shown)


tuky191 (commented):
## Code Review: PR #84 — Identity: face embedding as a gate (#83 phase 1)

229 tests pass. +47 −16, 2 files, 1 commit. Clean and focused.

### No issues

The face gate is correctly placed in the cascade — after the embedding floor (so OSNet/MobileNetV3 screening runs first) and before geometric/label overrides (so neither `geometricOverride` nor `labelOverride` can rescue a face-rejected candidate). This is the right veto semantics: face says different person → dead, regardless of body match.

Gate condition is properly scoped: `lockedIsPerson && candidateIsPerson && lockedFaceEmbedding != null && candidate.faceEmbedding != null`. Only fires for person-to-person with face data on both sides. Candidates without face data (BlazeFace failed, occluded, distant) fall through to the existing OSNet path unchanged.

`FACE_FLOOR = 0.40` is well-justified: the on-device data shows same-person face sim ≥ 0.5, different-person ≤ 0.3. The 0.06-0.26 range in the test session is well below the floor. The gap is wider than OSNet's (which has wrong-person at 0.55-0.70), confirming face is the cleaner signal.

Test updates are semantically correct: both changed from "scores higher" to "gate rejects" — matching the behavioral shift from scoring weight to hard gate. The `face gate vetoes good body` test properly constructs a case where OSNet passes (~0.99 reId) but face fails (~0.10 face) to verify the veto.

Good to merge.

PR #87, opened 2026-04-26
Synthetic stock-video tests for person identity (#62)

### LINKED ISSUE #62: Follow-ups to #49: detector-side perf + unlocked GLES3 wins

tuky191:
## Summary

Closes #62. Sources four Pexels CC0 stock videos that target person-tracking failure modes our existing fixtures didn't cover, and adds an EfficientDet-Lite2 helper for picking lock boxes without a GUI annotator.

| Test | Source | Length | Scenario | First-run result |
|---|---|---|---|---|
| `two_men_forest` | Pexels 4830416 | 17s | Two men, similar earth-tone clothing | 73% / 3 reacqs / face gate engaged |
| `two_men_suits` | Pexels 8915048 | 8s | Two men, near-identical black suits | 100% / 0 reacqs |
| `person_distance` | Pexels 5320110 | 14s | Single subject, ~6× scale change | 100% / 1 reacq |
| `crowd_street` | Pexels 12699538 | 12s | Six person detections in lock frame | 96% / 1 reacq |

## What this exercises

- **Color-histogram-can't-discriminate**: `two_men_forest` and `two_men_suits` put two similarly-clothed people in the same scene. Hair color and face are the only real discriminators — directly tests the face gate (#84) and OSNet body re-id paths.
- **Scale tolerance**: `person_distance` grows the locked subject ~6× over the clip. Gallery embeddings learned from a small initial crop have to keep matching as the subject fills the frame. The first run engaged the embedding-override path cleanly.
- **High-density same-category**: `crowd_street` has six person detections in the lock frame alone. Tests reacquisition through occlusion when many candidates pass the label gate simultaneously. Drop rate was 84% (per-frame embedding cost scales with persons in scene).

## Honest caveat

Three of the four tests probe identity (A vs B) but assert via `wrongCategoryReacqs(PERSON_LABELS)` — both subjects are 'person', so an A→B identity swap is invisible at the assertion layer. Same blind spot as `boy_indoor_wife_swap`. The reacquire metrics on these tests are still useful as a smoke test (catastrophic mistracking would push tracking rate near zero), but for true ground-truth verification we'd need per-frame-annotated labels. Tracked separately.

## Helper: `tools/synthetic_test_videos/detect_persons.py`

Runs the app's EfficientDet-Lite2 model on a still frame and prints normalized bboxes. Used to pick the `crowd_street` lock target — manual eyeballing wouldn't have produced a clean box in that scene. Generally useful for future fixture setup.

## Convention

`.mp4` files are gitignored (per existing `*.mp4` rule). JSON specs are committed; each carries `_source` URL and `_lockTarget` rationale so the fixture is reproducible — download the video from Pexels, push to `/sdcard/Android/data/com.haptictrack/files/test_videos/`, run.

## Test plan

- [x] Each test passes individually on device
- [ ] Full suite re-run after merge to confirm no interference with existing tests
- [ ] (deferred) Ground-truth annotation for two_men_*, crowd_street so A→B swaps fail loudly

### DIFF
--- a/app/src/androidTest/java/com/haptictrack/tracking/VideoReplayTest.kt
+++ b/app/src/androidTest/java/com/haptictrack/tracking/VideoReplayTest.kt
@@ -334,6 +334,27 @@ class VideoReplayTest {
             "totalFrames=${result.totalFrames}")
     }
 
+    // person_distance: synthetic stock-video test (Pexels 5320110, cottonbro studio,
+    // CC0). 13.6s — single shirtless man with distinctive tattoos walks from
+    // mid-distance toward camera, ending at close-up. Tests scale tolerance:
+    // gallery embeddings learned from a small crop (~12% of frame height at
+    // lock time) must continue to match as the subject grows ~6× over the clip.
+    // Adaptive size threshold and re-id at varied scale are exercised here.
+
+    @Test
+    fun person_distance_tracking_through_scale_change() {
+        val result = replayVideo("person_distance")
+
+        // Single-subject video — no wrong-person assertion is meaningful here.
+        // Primary signal is whether tracking survives the scale change. A timeout
+        // means the gallery couldn't bridge the small→large transition.
+        assertFalse("Should not timeout", result.timedOut)
+
+        Log.i(TAG, "person_distance: trackingRate=${result.trackingRate}% " +
+            "reacqs=${result.reacquisitions} losses=${result.losses} " +
+            "totalFrames=${result.totalFrames}")
+    }
+
     // --- Replay infrastructure ---
 
     data class ReplayEvent(
--- a/app/src/androidTest/java/com/haptictrack/tracking/VideoReplayTest.kt
+++ b/app/src/androidTest/java/com/haptictrack/tracking/VideoReplayTest.kt
@@ -355,6 +355,32 @@ class VideoReplayTest {
             "totalFrames=${result.totalFrames}")
     }
 
+    // crowd_street: synthetic stock-video test (Pexels 12699538, CC0). 12.2s
+    // portrait shot of a busy market street — six detected persons in the lock
+    // frame alone. Lock target picked via EfficientDet-Lite2 detection (the
+    // same model the app uses) so the box is what the production detector
+    // would produce. Tests reacquisition through occlusion and discrimination
+    // among many same-category candidates at once.
+
+    @Test
+    fun crowd_street_no_wrong_category_lock() {
+        val result = replayVideo("crowd_street")
+
+        // The dense-multi-person scene means an "A→B" person swap is invisible
+        // to PERSON_LABELS. Primary signals: no timeout (the lock should be
+        // reacquired through brief occlusions) and tracking rate that doesn't
+        // collapse (catastrophic mistracking would push it near zero).
+        assertFalse("Should not timeout", result.timedOut)
+
+        val wrong = result.wrongCategoryReacqs(PERSON_LABELS)
+        assertTrue("Should never reacquire non-person (got: ${wrong.map { "${it.label}@F${it.frame}" }})",
+            wrong.isEmpty())
+
+        Log.i(TAG, "crowd_street: trackingRate=${result.trackingRate}% " +
+            "reacqs=${result.reacquisitions} losses=${result.losses} " +
+            "totalFrames=${result.totalFrames}")
+    }
+
     // --- Replay infrastructure ---
 
     data class ReplayEvent(
--- a/app/src/androidTest/java/com/haptictrack/tracking/VideoReplayTest.kt
+++ b/app/src/androidTest/java/com/haptictrack/tracking/VideoReplayTest.kt
@@ -279,6 +279,61 @@ class VideoReplayTest {
             result.trackingRate >= 25)
     }
 
+    // two_men_forest: synthetic stock-video test (Pexels 4830416, cottonbro studio,
+    // CC0). Two men with similar earth-tone outdoor clothing walk down a path.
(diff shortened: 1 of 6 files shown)


tuky191 (commented):
## Code Review: PR #87 — Synthetic stock-video tests for person identity (#62)

232 unit tests pass. +182, 6 files. Test fixtures + helper tool, no production code changes.

### No issues

Clean test fixture addition. Four well-chosen Pexels CC0 videos targeting specific failure modes:

- **two_men_forest** / **two_men_suits**: Color histogram can't discriminate — probes face gate + OSNet. Honest caveat that `PERSON_LABELS` assertion can't catch A→B identity swaps (same blind spot documented).
- **person_distance**: Scale tolerance (~6× growth) — tests gallery robustness across crop sizes.
- **crowd_street**: High-density (6 persons) — tests discrimination among many same-category candidates. 60fps spec is a nice detail (higher frame rate = more frames to process).

JSON specs include `_source` URL, `_lockTarget` rationale, and `_caveat` documentation — excellent for reproducibility. MP4s correctly gitignored per existing convention.

`detect_persons.py` helper is minimal and useful: runs the same EfficientDet model as the app to pick lock boxes. Clean single-file script.

No tracking-rate assertions yet (marked "Baseline TBD") — appropriate for first commit. Assertions are smoke-test level: no timeout, no wrong-category reacquire.

Good to merge.

PR #88, opened 2026-04-26
Ground-truth identity validation for VideoReplayTest

tuky191:
## Summary

Adds an independent ground-truth pipeline so multi-person tests can detect A→B identity swaps that `wrongCategoryReacqs(PERSON_LABELS)` misses (because both subjects are 'person' to the assertion layer — the same blind spot as `boy_indoor_wife_swap`'s known bug).

## Pieces

**`tools/generate_ground_truth.py`** — YOLOv8 person detection + ByteTrack multi-object tracking on the full video. Picks the lock track at `lockFrame` by best IoU to the spec's lockBox, then follows that track ID through the rest of the video. Emits:
- `<video>.gt.json` — per-frame normalized bbox (or null when ByteTrack reports lost)
- `<video>_gt_preview/` — sampled preview frames for human visual review

**`VideoReplayTest.kt`** —
- `ReplayEvent` gains `videoFrame: Int?` + `box: FloatArray?`, captured at LOST/REACQUIRE/TIMEOUT
- `GroundTruth.loadOrNull(file)` — reads sidecar JSON if present, returns null otherwise (no-op when absent — **fail-open**)
- `ReplayResult.wrongIdentityReacqs(iouThreshold)` — for each REACQUIRE, compares its bbox to GT at that videoFrame; flags when IoU < threshold; skips frames where GT is null (don't enforce identity where GT itself is uncertain)
- `boy_indoor_wife_swap_reacquires_correctly` is the first consumer

## Why YOLOv8 + ByteTrack as ground truth

It's an entirely different model family from the app stack (EfficientDet + MobileNetV3 embedding + OSNet body re-id + ViT visual tracker). When the app stack swaps subjects and ByteTrack stays correct, GT catches it.

## Honest limitation

ByteTrack itself can drift on long heavy occlusions or repeated subject-crossings. On `boy_indoor_wife_swap` it stayed correct through frame ~1080, then swapped to the wife. Mitigated by sampling preview frames and manually trimming post-drift annotations to null — the test loader silently skips null frames.

## First run

`boy_indoor_wife_swap_reacquires_correctly` with 1094 confident GT frames (frames 1100+ trimmed):
- Test passes.
- All 6 reacquires in the GT-confident range had IoU 0.88–0.96 with GT — engine tracked the boy correctly on those frames.
- 6 post-1100 reacquires are skipped (untrusted GT). Validating those needs manual annotation or a more robust tracker.

## Test plan

- [x] `compileDebugAndroidTestKotlin` clean
- [x] `testDebugUnitTest` passes
- [x] `boy_indoor_wife_swap_reacquires_correctly` passes on device with GT loaded
- [ ] Apply to `person_playground` and other multi-person fixtures (follow-ups in this PR)
- [ ] Will conflict with PR #87 (`feat/synthetic-test-videos-62`) on `VideoReplayTest.kt`; rebase the later-merging branch.

### DIFF
--- a/app/src/androidTest/java/com/haptictrack/tracking/VideoReplayTest.kt
+++ b/app/src/androidTest/java/com/haptictrack/tracking/VideoReplayTest.kt
@@ -139,9 +139,20 @@ class VideoReplayTest {
         assertTrue("Should never reacquire non-person (got: ${wrong.map { "${it.label}@F${it.frame}" }})",
             wrong.isEmpty())
 
+        // Identity check: the playground often has multiple people at varying
+        // distances. Earlier runs of this test showed up to 9 reacquires —
+        // some of which could have jumped to a passing person rather than the
+        // locked subject. GT skips frames where ByteTrack itself was lost.
+        val swapped = result.wrongIdentityReacqs(iouThreshold = 0.3f)
+        assertTrue(
+            "Should never reacquire onto a different person. " +
+                "Mismatches: ${swapped.map { "frame ${it.videoFrame} box=${it.box?.toList()}" }}",
+            swapped.isEmpty()
+        )
+
         Log.i(TAG, "person_playground: trackingRate=${result.trackingRate}% " +
             "reacqs=${result.reacquisitions} losses=${result.losses} " +
-            "totalFrames=${result.totalFrames}")
+            "totalFrames=${result.totalFrames} gt=${result.groundTruth?.annotatedFrameCount ?: 0}")
     }
 
     @Test
--- a/app/src/androidTest/java/com/haptictrack/tracking/VideoReplayTest.kt
+++ b/app/src/androidTest/java/com/haptictrack/tracking/VideoReplayTest.kt
@@ -168,9 +168,21 @@ class VideoReplayTest {
         assertTrue("Should never reacquire non-person (got: ${wrong.map { "${it.label}@F${it.frame}" }})",
             wrong.isEmpty())
 
+        // Identity check via ground truth (sidecar .gt.json from
+        // tools/synthetic_test_videos/generate_ground_truth.py). Empty when
+        // GT is absent — we don't fail closed on missing ground truth, but
+        // when it exists we enforce that REACQUIRE bboxes match the lock
+        // subject's GT bbox (IoU >= 0.3) on frames where GT is confident.
+        val swapped = result.wrongIdentityReacqs(iouThreshold = 0.3f)
+        assertTrue(
+            "Should never reacquire onto a different person (son↔wife swap). " +
+                "Mismatches: ${swapped.map { "frame ${it.videoFrame} box=${it.box?.toList()}" }}",
+            swapped.isEmpty()
+        )
+
         Log.i(TAG, "boy_indoor_wife_swap: trackingRate=${result.trackingRate}% " +
             "reacqs=${result.reacquisitions} losses=${result.losses} " +
-            "totalFrames=${result.totalFrames}")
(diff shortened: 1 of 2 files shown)


tuky191 (commented):
## Code Review: PR #88 — Ground-truth identity validation for VideoReplayTest

232 unit tests pass. +360 −8, 2 files. Addresses the documented blind spot where `wrongCategoryReacqs(PERSON_LABELS)` can't detect person→person identity swaps.

### Medium

**1. Unsafe cast in GroundTruth.loadOrNull** (`VideoReplayTest.kt:113`)

```kotlin
val arr = box as org.json.JSONArray
```

If the JSON `"box"` field is present but not an array (e.g., malformed GT file), this throws `ClassCastException`. Should use `a.optJSONArray("box")` instead, which returns null on type mismatch:

```kotlin
val arr = a.optJSONArray("box") ?: continue
```

This also eliminates the separate `if (box == JSONObject.NULL) continue` check since `optJSONArray` handles null values.

### Low

**2. `wrongIdentityReacqs` IoU threshold at 0.3 is lenient** (`VideoReplayTest.kt:147`)

IoU 0.3 means only 30% overlap between the app's reacquire box and GT box is required to count as "same subject." Two overlapping persons (common in crowd/indoor scenes) could have IoU > 0.3 while being different people. The first-run data (0.88-0.96 IoU on good matches) suggests 0.5 would work and catch more edge cases. The 0.3 default is conservative-safe for now.

**3. `generate_ground_truth.py` model path fallback** (`generate_ground_truth.py:301-309`)

The model path resolution tries relative-to-script and relative-to-repo-root but doesn't try CWD. If invoked from the repo root with `--model yolov8n.pt` and the model is in CWD, it still finds it (argparse gives a relative path that Python resolves against CWD). But the candidates list skips CWD explicitly, which could confuse someone reading the code. Minor.

- **Fail-open design is correct**: `GroundTruth.loadOrNull` returns null when GT file is missing → `wrongIdentityReacqs` returns empty list → no assertion fires. Tests don't fail when GT isn't pushed to device. GT adds value when present, doesn't break anything when absent.

- **ByteTrack limitations honestly documented**: The generator script's docstring and the test comments both note that ByteTrack can drift on long occlusions, and that post-drift frames should be manually nulled out. The `boy_indoor_wife_swap` case (frames 1100+ trimmed) demonstrates the workflow.

- **Chair and flowerpot tests correctly skip GT**: Comments explain why — YOLOv8 only detects the chair in ~14% of frames and the flowerpot in ~7%, so GT would silently no-op on most reacquires.

- **ReplayEvent extended cleanly**: `videoFrame: Int?` and `box: FloatArray?` are nullable with defaults, so existing event construction (without these fields) continues to work. REACQUIRE events capture `lastKnownBox` at event time — correct for GT comparison.

- **GT is an independent model family**: YOLOv8 + ByteTrack vs the app's EfficientDet + MobileNetV3 + OSNet + ViT. When the app swaps subjects and ByteTrack stays correct, the test catches it. Good orthogonality.

Good to merge after the cast fix (#1).

tuky191 (commented):
## Follow-up review (commit 41fe034)

| # | Finding | Status |
|---|---------|--------|
| **1 (Med)** | Unsafe `box as JSONArray` cast | **Fixed.** Replaced with `a.optJSONArray("box") ?: continue` + `arr.length() < 4` guard. Handles null, JSONObject.NULL, and non-array types cleanly. |
| **2 (Low)** | IoU threshold at 0.3 too lenient | **Fixed.** Default raised to 0.5 with empirical justification (correct matches at 0.88-0.96). Call sites updated to use default. |
| **3 (Low)** | Model path resolution skips CWD | **Fixed.** CWD added as first candidate, resolved via `.resolve()`. |

Good to merge.

PR #89, opened 2026-04-26
Scene face-body memory for asymmetric identity gating (#83 phase 2)

tuky191:
**Status: draft — awaiting on-device verification.** APK builds and installs cleanly on master + this branch; the assertion logic is unit-tested but the live-camera scenario hasn't been exercised yet.

## Summary

Closes part of #83. Builds a bounded paired (face, body) memory of non-lock persons observed during VT-confirmed tracking, used asymmetrically at gate time to veto wrong-person reacquires that #67 (OSNet gate) and #84 (face gate) miss.

## Mechanism

During VT-confirmed tracking, every 5 confirmed frames, for each non-lock `person` detection: compute both face (MobileFaceNet) and body (OSNet) embeddings and store the pair temporally aligned. At gate time:

- **Face-known path** — candidate has a face. Compare to stored other-faces. If the best stored-face match exceeds `SCENE_FACE_MATCH_FLOOR=0.40` AND beats the lock-face match → reject. "I recognize this face from earlier; not the lock."
- **Body-known path** — candidate has no face. Compare body to stored other-bodies. If the best stored-body match beats the lock-body match by `SCENE_BODY_VETO_MARGIN=0.10` → reject. The earlier face match (when the body was observed alongside a face) carries the identity weight forward.

Same memory, two failure modes covered.

## Invariants

- Only accumulate during VT-confirmed tracking (during search we don't know lock vs other).
- Filter near-lock pairs: `addScenePersonPair` rejects entries with body sim ≥ 0.85 to lock gallery OR face sim ≥ 0.6 to lock face — otherwise duplicate detections of the lock would get stored as "an other".
- Bounded ring of 16 entries; oldest evicted on overflow.
- Reset on lock/clear.

## Cost

Per gate: O(memory_size × 2) cosine sims ≈ 32 dot products per candidate. < 0.5ms on Adreno 740. Negligible.

## Test plan

- [x] Unit tests pass (238 total, 6 new for the memory mechanism)
- [x] APK builds cleanly on rebased branch
- [ ] **On-device verification** — the canonical scenario:
  1. Lock on person A (face visible at lock)
  2. Track ~5s with person B also in frame (face visible to B too)
  3. Pan away to lose lock
  4. Pan back so only B is visible → expect `SCENE_FACE_VETO` log line, no false reacquire
  5. Pan back to A → reacquires correctly

Filter command:
```
adb logcat -s "Reacq" | grep -E "SCENE_FACE_VETO|SCENE_BODY_VETO|REACQUIRE"
```

### DIFF
--- a/app/src/main/java/com/haptictrack/tracking/ObjectTracker.kt
+++ b/app/src/main/java/com/haptictrack/tracking/ObjectTracker.kt
@@ -447,12 +447,19 @@ class ObjectTracker(
                         // Collect scene negatives every 5 confirmed frames.
                         // tracked has screen-space boxes but bitmap is rotated,
                         // so convert back to rotated-image space before cropping.
-                        // Also build paired face+body memory for non-lock persons
-                        // (#83 phase 2) — at gate time the engine compares candidates
-                        // against this memory asymmetrically (face match ⇒ veto,
-                        // body match exceeding lock match ⇒ veto when face missing).
+                        // Scene negatives + paired face+body memory (#83 phase 2).
+                        // Both run on this 5-confirmed-frame cadence. We tried
+                        // slowing scene-pair collection to every 15 frames in
+                        // response to the perf review (face+body inference is
+                        // ~16-20ms per non-lock person), but on real boy_indoor
+                        // footage that cut accumulation to zero pairs across the
+                        // full clip — short multi-person lock windows didn't see
+                        // the wife enough at %15 to learn her. The mechanism
+                        // needs the data more than the hot path needs the ms;
+                        // keeping %5 and noting async batching as a follow-up
+                        // optimization once we can measure real device fps cost.
                         if (vtConfirmedFrames % 5 == 0) {
-                            val isPersonLock = reacquisition.lockedIsPerson
+                            val collectScenePair = reacquisition.lockedIsPerson
                             for (det in tracked) {
                                 if (det.id == reacquisition.lockedId) continue
                                 val rotBox = mapToRotated(det.boundingBox.left, det.boundingBox.top,
@@ -460,9 +467,7 @@ class ObjectTracker(
                                 val negEmb = appearanceEmbedder.embed(bitmap, rotBox)
                                 if (negEmb != null) reacquisition.addSceneNegative(negEmb)
 
-                                // For person-vs-person: link face+body if both present.
-                                // Same time-aligned pair makes the asymmetric veto sound.
-                                if (isPersonLock && det.label == "person") {
+                                if (collectScenePair && det.label == "person") {
                                     val faceEmb = faceEmbedder.embedFace(bitmap, rotBox)
                                     val bodyEmb = personReId.embed(bitmap, rotBox)
                                     if (faceEmb != null && bodyEmb != null) {
--- a/app/src/main/java/com/haptictrack/tracking/ObjectTracker.kt
+++ b/app/src/main/java/com/haptictrack/tracking/ObjectTracker.kt
@@ -447,13 +447,28 @@ class ObjectTracker(
                         // Collect scene negatives every 5 confirmed frames.
                         // tracked has screen-space boxes but bitmap is rotated,
                         // so convert back to rotated-image space before cropping.
+                        // Also build paired face+body memory for non-lock persons
+                        // (#83 phase 2) — at gate time the engine compares candidates
+                        // against this memory asymmetrically (face match ⇒ veto,
+                        // body match exceeding lock match ⇒ veto when face missing).
                         if (vtConfirmedFrames % 5 == 0) {
+                            val isPersonLock = reacquisition.lockedIsPerson
                             for (det in tracked) {
                                 if (det.id == reacquisition.lockedId) continue
                                 val rotBox = mapToRotated(det.boundingBox.left, det.boundingBox.top,
                                     det.boundingBox.right, det.boundingBox.bottom, deviceRot)
                                 val negEmb = appearanceEmbedder.embed(bitmap, rotBox)
                                 if (negEmb != null) reacquisition.addSceneNegative(negEmb)
+
+                                // For person-vs-person: link face+body if both present.
+                                // Same time-aligned pair makes the asymmetric veto sound.
+                                if (isPersonLock && det.label == "person") {
+                                    val faceEmb = faceEmbedder.embedFace(bitmap, rotBox)
+                                    val bodyEmb = personReId.embed(bitmap, rotBox)
+                                    if (faceEmb != null && bodyEmb != null) {
+                                        reacquisition.addScenePersonPair(faceEmb, bodyEmb)
+                                    }
+                                }
                             }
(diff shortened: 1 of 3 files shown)


tuky191:
## On-device verification via `boy_indoor_wife_swap_reacquires_correctly`

Ran the existing replay test on this branch — the documented son↔wife scene exercises both veto paths.

**Result: PASS** (98.7s wall time, comparable to master)

| Metric | Master baseline | This branch |
|---|---|---|
| Tracking rate | 79% | 78% |
| Reacquires / losses | 12 / 12 | 12 / 12 |
| GT identity mismatches (#88) | 0 | 0 |

Phase-2 fires actively during the test — both paths exercised:

**SCENE_FACE_VETO** (face-known path):
```
21:07:19  sceneFace=0.752 > lockFace=0.601 (reIdSim=0.839)   ×9 in cluster
21:08:05  sceneFace=0.493 > lockFace=0.488 (reIdSim=0.779)   ×6 in cluster
```

**SCENE_BODY_VETO** (body-known path, candidate face missing):
```
21:08:07  sceneBody=0.784 > lockBody=0.654 + 0.10 margin     ×6 in cluster
          sceneBody=0.760 > lockBody=0.635 + 0.10
          sceneBody=0.764 > lockBody=0.619 + 0.10
```

Body-veto path engages exactly when face is missing — asymmetric design working as intended.

## What this confirms

- ✅ Mechanism fires on realistic multi-person input (sim values plausible, no NaNs/zeros)
- ✅ Both face-known and body-known paths exercised
- ✅ No false positives — same reacquire count as master, GT identity check still passes (no legitimate reacquires blocked)

## What it doesn't fully prove

We can't tell from this test alone whether each individual veto was *correct* — i.e., that the rejected candidate truly was the wife and not the son. For that we'd want per-frame GT for both subjects. But the engine still reacquires the right subject after the vetoes (GT identity check at IoU 0.5 passes), so on aggregate the mechanism is at least not making things worse.

The full hands-on scenario (live two-person, point-camera-back-at-other-person) is still a good follow-up but isn't blocking review/merge.

tuky191 (commented):
## Code Review: PR #89 — Scene face-body memory for asymmetric identity gating (#83 phase 2)

238 tests pass (6 new). +283 −3, 3 files. Draft — awaiting on-device verification.

### Medium

**1. Face+body computation adds ~30-50ms per non-lock person every 5 VT-confirmed frames** (`ObjectTracker.kt:460-472`)

During VT-confirmed tracking, every 5 frames, for each non-lock person detection: `faceEmbedder.embedFace` (~8ms) + `personReId.embed` (~8-12ms). This runs on the processing thread synchronously (not the embedding executor). With 2-3 non-lock persons visible, that's ~40-75ms added every 5 frames — a significant periodic spike on the hot path.

This is in addition to the existing `appearanceEmbedder.embed` call (scene negatives) that also runs every 5 frames. Consider batching face+body collection onto the embedding executor alongside scene negatives, or reducing the collection frequency for face-body pairs (e.g., every 15 frames, matching gallery accumulation cadence).

**2. Face path (a) fires even when lock face barely passes FACE_FLOOR** (`ReacquisitionEngine.kt:782`)

The condition is `sceneFaceMatch >= SCENE_FACE_MATCH_FLOOR && sceneFaceMatch > lockFaceSim`. If lock face sim is 0.41 (barely passed FACE_FLOOR at 0.40) and scene face match is 0.42, the veto fires even though both are marginal. The `>` operator has no margin — a tiny numerical difference between lock-face and scene-face triggers a hard reject.

Consider adding a small margin (e.g., `sceneFaceMatch > lockFaceSim + 0.05f`) so the veto only fires when the scene face is meaningfully better than the lock face, not on measurement noise.

### Low

**3. `addScenePersonPair` body filter uses gallery (MobileNetV3), not OSNet** (`ReacquisitionEngine.kt:262-263`)

The near-lock body filter computes `bestGallerySimilarity(body, _embeddingGallery)` where `_embeddingGallery` contains MobileNetV3 embeddings and `body` is an OSNet embedding. These are different embedding spaces (1280-dim vs 512-dim) — the cosine similarity between them is meaningless.

Wait — checking the code: `body` is the OSNet output from `personReId.embed()`. The gallery is MobileNetV3 from `appearanceEmbedder`. These ARE different dimensions. But `cosineSimilarity` returns 0 for mismatched sizes (line 26 of EmbeddingUtils: `if (a.size != b.size) return 0f`). So the filter silently falls through to the face check only. The body-based near-lock filter is dead code — it will never reject.

The face filter at `cosineSimilarity(lockedFaceEmbedding, face) >= 0.6f` works correctly (both are MobileFaceNet, same dimension). So near-lock filtering relies entirely on the face check. If the lock's face is null (locked when face wasn't visible), NO near-lock filtering fires, and any non-lock person pair gets stored — including duplicate detections of the lock person.

- **Gate cascade placement is correct**: Scene memory veto at line 765 fires AFTER the face gate (candidate already proved it's not a lock-face mismatch) but BEFORE geometric/label overrides (strong appearance can't rescue a scene-vetoed candidate). Hard `return null` semantics are appropriate for an identity veto.

- **Asymmetric design is sound**: Face path (a) covers the case where the candidate's face is visible. Body path (b) covers the case where it isn't — the earlier face-body link carries identity forward. The `!hasFaceGate` guard on path (b) correctly prevents both paths from firing on the same candidate.

- **State lifecycle is clean**: `_sceneFacePairs.clear()` in both `lock()` (line 363) and `clear()` (line 454). `MAX_SCENE_FACE_PAIRS = 16` with FIFO eviction. No leakage across sessions.

- **Tests cover the key scenarios**: Face veto, body veto, genuine lock pass-through, near-lock body filter, near-lock face filter, and ring bound. Good coverage for a new mechanism.

- **`ScenePersonPair` uses `FloatArray` correctly**: `copyOf()` on both face and body in `addScenePersonPair` prevents aliasing with the caller's buffers.

Good to merge after on-device verification and addressing the dimension mismatch in finding #3.

tuky191 (commented):
## Follow-up review (commit 4d601ac)

239 tests pass (+1 new margin test).

| # | Finding | Status |
|---|---------|--------|
| **1 (Med)** | Face+body computation adds 30-50ms on hot path | **Documented.** Comment explains 15-frame cadence was tried but produced zero pairs on real footage. Async batching noted as follow-up. Pragmatic call — the mechanism needs the data. |
| **2 (Med)** | Face veto fires on measurement noise (no margin) | **Fixed.** `SCENE_FACE_VETO_MARGIN = 0.05f` added. Veto now requires `sceneFaceMatch > lockFaceSim + 0.05f`. New test verifies margin prevents false veto on near-equal scores. |
| **3 (Low)** | Body filter compared OSNet against MobileNetV3 gallery (dead code) | **Fixed.** Now compares `cosineSimilarity(lockedReIdEmbedding, body)` — same OSNet space. Test vectors updated to matching dimensions. Log message added for rejected near-lock pairs. |

Good to merge (after on-device verification).

Issue #96, opened 2026-04-27
VideoReplayTest hangs after 5–6 tests on shared tracker (#92 follow-up)


tuky191:
Found while collecting Phase 0 baseline (#92). Doesn't block #92 shipping, but blocks Phase 1 PRs that need a clean full-suite re-baseline.

## Symptom

Running the full \`VideoReplayTest\` suite via \`am instrument\`:

\`\`\`
adb shell am instrument -w com.haptictrack.test/androidx.test.runner.AndroidJUnitRunner
\`\`\`

…hangs reliably on the second or third test. App process stays alive, RSS stays flat, but CPU usage drops to 0%. \`adb logcat\` produces no further \`VideoReplayTest\`/\`STFrameReader\` lines from the hung test. \`dumpsys thermalservice\` reports thermal status 0 (no throttling), so the device is fine.

Reproduced twice on 2026-04-27:

| Run | Tests completed | Where it hung |
|---|---|---|
| Synchronous-on-pipeline audit | man_desk, person_playground (slow), mouse_desk, two_men_forest, then man_desk (#2) hung at 17 min in | VT-confirmed loop on man_desk_camera_swing_tracking_rate |
| Async audit executor (\`auditExecutor\`) | man_desk, then person_playground hung at 11+ min | unclear; person_playground GL pipeline is unrelatedly slow on the 277MB file |

Single \`am instrument -e class …#methodName\` invocations *mostly* complete cleanly (~30–60s each) but two of four single-test isolated runs in the same baseline session also hung at the 3-min watchdog (boy_indoor_wife_swap and crowd_street).

## Likely candidates

The hang is probably shared-state buildup on \`sharedTracker\` (the @BeforeClass-instantiated ObjectTracker reused across all tests in the class). Specifically:

1. **GPU context leaks.** Each TFLite GPU embedder + the segmenter holds a \`GpuDelegate\`. Per the Adreno 740's documented limits, repeated GPU work without proper cleanup can saturate the available contexts. The audit doubles the segmenter call rate by running MNV3 on the locked object every 5 confirmed frames in addition to the existing template self-verification.

2. **Embedder synchronization.** \`PersonReIdEmbedder.embedFromCrop\` and \`FaceEmbedder.embedFaceFromCrop\` are both \`@Synchronized\`. With the audit on a separate thread, two threads can contend for the embedder monitors. A bug in MediaPipe's \`ImageEmbedder\` or the TFLite Java wrapper might deadlock on this contention.

[...]

Issue #94, opened 2026-04-27
Phase 1B (#91): thread masked crop through OSNet (single segmentation pass)


tuky191:
Parent: #91 — embedding input audit.
Depends on: Phase 0 baseline.

## Problem

Asymmetry: \`AppearanceEmbedder.embedAndCrop\` masks background via \`ObjectSegmenter.segmentAndCrop\` before MNV3, but \`PersonReIdEmbedder.embed()\` (\`PersonReIdEmbedder.kt:69\`) calls \`cropNormalized\` with the raw bbox.

The gates we lean on most (OSNet body re-id #67, scene memory body match #89) embed wallpaper, sofa, and the person standing behind the lock. The gate we lean on least (MNV3, ~10% of scoring) gets the cleanest input.

## Change

Modify \`ObjectTracker\` so each person detection runs segmentation **once**, sharing the masked crop with both MNV3 and OSNet:

- \`AppearanceEmbedder.embedAndCrop\` already returns \`EmbedResult(embedding, maskedCrop)\` — extend its consumers in \`ObjectTracker\` (lines 241, 442, 471–472, 741, 1019) to feed \`personReId.embedFromCrop(maskedCrop)\` when the masked crop is available
- Fallback: when segmentation returns null (>50% area, low fg %, segmenter failure), OSNet keeps current raw-crop path
- Face stays on raw person crop — masking shoulders before BlazeFace risks suppressing face detection. Phase 0 will produce evidence for/against this; revisit if Phase 0 shows face detection survives masking.

## Measurement (vs Phase 0 baseline)

- OSNet same-person sim p50 should **rise** on multi-person videos
- OSNet different-person sim should stay roughly flat — wider separation is the win, not just shifted up
- Wrong-identity reacqs: drop on \`boy_indoor_wife_swap\`, \`person_playground\`
- Frame-budget impact: zero, because we're sharing a segmentation pass that already runs for MNV3

## Risks

- Segmenter rejects the mask (>50% crop area or fg<5% / fg>95% per \`ObjectSegmenter.kt:135-138\`) → OSNet uses raw crop. Don't make masking required.
- The masked crop dimensions equal the bbox (post-pad), not OSNet's 256×128. Resize policy: same as today (let the embedder handle its own resize).

## Code refs

- \`PersonReIdEmbedder.kt:68-75,82-100\` — already supports \`embedFromCrop\`; just route the masked one in
[...]

Issue #93, opened 2026-04-27
Phase 1A (#91): face alignment via BlazeFace keypoints


tuky191:
Parent: #91 — embedding input audit.
Depends on: Phase 0 baseline (for measurable comparison).

Highest single-fix ROI per #91. Affects three gates: face gate (#84), scene face memory (#89), tentative confirmation (#52).

## Problem

\`FaceEmbedder.embedFaceFromCrop\` (\`FaceEmbedder.kt:87\`) crops the BlazeFace bounding box and stretches it to 112×112. BlazeFace returns 6 keypoints (2 eyes, nose, mouth, 2 ears) — they're ignored.

Standard face-recognition pipelines (InsightFace, ArcFace, MobileFaceNet) do a 5-point similarity transform: source keypoints → fixed canonical positions, applied via affine. Without alignment the model sees tilted, off-center, partial faces — embedding noise that no margin tweak fixes. Documented loss: 10–20% accuracy on face verification benchmarks.

## Change

Add \`align5pt(faceCrop, keypoints)\` to \`FaceEmbedder\`:

- Use OpenCV \`getAffineTransform\` + \`warpAffine\` to canonical 112×112
- Source: 5 BlazeFace keypoints (2 eyes, nose, mouth corners — derived from the 6 BlazeFace points)
- Target: ArcFace canonical positions for 112×112
- Replace bbox-crop + stretch with crop → align → 112×112
- Fallback: if keypoint detection is poor (low confidence, missing keypoints), use the current bbox path

## Measurement (vs Phase 0 baseline)

- Self-similarity p50 for MobileFaceNet should **rise** (less alignment jitter)
- Same-person face cosine between gallery and confirmed VT-tracked candidate should rise
- Wrong-person face cosine on \`boy_indoor_wife_swap\` (boy gallery vs wife candidates) should **not** rise — or rise less than same-person
- Wrong-identity reacqs on \`boy_indoor_wife_swap\` and \`person_playground\` (GT-checked per #88): drop or stay zero

## Risks

- Alignment matrix breaks if keypoints are bad (small/blurry faces) — fallback path required
- BlazeFace keypoint API: verify 6-keypoint output is exposed on the MediaPipe Tasks Vision \`FaceDetector\` we use

## Verification

Before/after composite PNGs (from Phase 0 instrumentation) showing the same face crop pre- and post-alignment, plus the delta table on the same replay videos.

## Code refs

[...]

Issue #92, opened 2026-04-27
Phase 0 (#91): baseline embedding instrumentation + crop debug capture


tuky191:
Parent: #91 — embedding input audit.

The first PR in the audit. No embedder behavior changes — this exists to give every subsequent fix a measured baseline plus visible crops to eyeball.

### 1. Visual crop capture
New \`CropDebugCapture\` (sibling of \`DebugFrameCapture\`):

- On LOCK: write per-detection composite PNG showing
  - full frame with bbox drawn
  - raw bbox crop (before any resize)
  - segmenter mask visualization (foreground green-tinted)
  - masked crop returned by \`ObjectSegmenter\`
  - MNV3 input (post-resize 224×224, pre-normalization)
  - OSNet input (post-resize 256×128, pre-normalization)
  - BlazeFace box drawn on person crop + the 6 keypoints
  - Face crop (post-resize 112×112, pre-normalization)
- During VT-confirmed tracking: same composite every 30 frames
- Cap at 20 composites per session, JPEG q=85, in \`session_*/crops/\`
- Gated by \`BuildConfig.DEBUG\` — zero cost in release

### 2. Self-similarity instrumentation
New \`EmbeddingStabilityLogger\`:

- Per embedder (MNV3, OSNet, MobileFaceNet), maintain a ring of last 32 embeddings of the locked object during VT-confirmed tracking
- Each new embedding: compute cosine to k=1, k=5, k=30 frames back; accumulate p10/p50/p90 + histogram
- On lock clear, write \`session_*/embedding_stability.json\`

### 3. Replay test aggregator
Extend \`VideoReplayTest\` to dump \`embedding_stability.json\` per video and print a comparison table at run end:

\`\`\`
video                          | embedder      | k=1   k=5   k=30
boy_indoor_wife_swap           | MNV3          | 0.94  0.89  0.81
                               | OSNet         | …
                               | MobileFaceNet | …
\`\`\`

## Exit criteria

- All existing replay tests still pass
- Composite PNGs render correctly on device for all six replay videos
- \`embedding_stability.json\` produced for each
- Markdown report committed (\`docs/embedding_audit_baseline.md\`) with:
  - the comparison table
  - 2–3 representative composites per video
  - notes on visible problems (loose boxes, distortion, background-heavy crops, misaligned faces)

## Why first

We can't compare fix-vs-baseline without a baseline. And the user gets to **see** what each embedder is fed — which may itself surface fixes we haven't thought of.

## Code refs

[...]

PR #97, opened 2026-04-27
Embedding-input audit: instrumentation + visible crops (#92)

tuky191:
## Summary

Phase 0 of #91. One-shot instrumentation that makes embedder inputs visible (composite JPEGs at LOCK + every 30 confirmed VT frames) and measures consecutive-frame embedding stability for MNV3, OSNet, and MobileFaceNet. Designed to give Phase 1 fixes (#93/#94/#95) a real baseline to beat.

**Default: off.** `CropDebugCapture.AUDIT_ENABLED = false` ships in this PR. The audit produces clean stability numbers when on, but the audit-thread `@Synchronized` embedder calls contend with the production search-mode pipeline enough to perturb other test assertions during a full suite run. So: flip the flag for one-shot collection, regenerate `docs/embedding_audit_baseline.md`, flip back. Workflow documented in the doc.

## What's in here

- **`CropDebugCapture`** — composite JPEG (full frame + bbox · raw bbox crop · masked crop · OSNet 256×128 input · BlazeFace bbox + 6 keypoints · MobileFaceNet 112×112 input). Three sample composites under `docs/embedding_audit_examples/`.
- **`EmbeddingStabilityLogger`** — per-embedder ring of locked-object embeddings, k=1 / 5 / 30 percentiles → `embedding_stability.json`. `VideoReplayTest` aggregates per-video summaries and prints a comparison table at `@AfterClass`.
- **Async via low-priority `auditExecutor`** — bitmap snapshot is the only cost on the processing thread. Bounded queue + DiscardOldest under backpressure.
- **`docs/embedding_audit_baseline.md`** — full table (11 tests, all 3 embedders) + 3 sample composites. Headlines:
  - **OSNet tightest**: k=1 p10 0.65–0.91, p50 0.91–0.97. #94 has the clearest room.
  - **MNV3 noisiest**: k=1 p10 collapses to **0.20** on `person_playground_tracking_rate` (segmenter rejects close-up bboxes; raw crop bleeds background).
  - **MobileFaceNet collapses in multi-person**: k=1 p10 = **0.29** on `person_playground`, **0.41** on `boy_indoor_wife_swap`. #93 is the highest-ROI fix.

## Test plan

- [x] 239 unit tests pass
- [x] Full `VideoReplayTest` suite runs end-to-end (16 tests, ~44 min) on Xiaomi Poco F4
- [x] With `AUDIT_ENABLED = true`, 14/16 tests pass; 2 failures are audit-induced contention regressions (`man_desk_camera_swing_tracking_rate`, `mouse_desk_rotation_reacquires_correctly`) — won't reproduce with default `false`
- [x] With `AUDIT_ENABLED = false` (the shipped default), full suite runs without audit perturbation

## Resolved during PR development

- **#96 (closed)** — what looked like a deadlock was a stale 277 MB test video on the device. Local checkout already had a 92 MB / 640 px version. After pushing the correct file, the suite completes cleanly. My watchdog interpreted "0% CPU during `MediaCodec.dequeueOutputBuffer` waits" as a hang — that's wrong; threads in S state during JNI codec waits are normal.


tuky191 (commented):
## Code Review — PR #97 (Embedding-input audit: instrumentation + visible crops)

**Scope:** Phase 0 of #91. Adds audit infrastructure (default-off) to visualize embedder inputs and measure frame-to-frame embedding stability. New files: `CropDebugCapture`, `EmbeddingStabilityLogger`. Debug helpers added to `AppearanceEmbedder`, `PersonReIdEmbedder`, `FaceEmbedder`. Integration into `ObjectTracker` for periodic audit sampling + drain on `clearLock`. `VideoReplayTest` aggregates per-video stability tables.

239 unit tests pass.

---

### Medium

**1. `captureExecutor` adds a third concurrent caller to synchronized embedder resources**
`CropDebugCapture` has its own `captureExecutor` thread (line 178). When `runAuditSample` on `auditExecutor` calls `cropDebugCapture.capture()`, the composite write runs on `captureExecutor`, which calls `appearanceEmbedder.debugMaskedCrop()` → `segmenter.segmentAndCrop()` (`@Synchronized`) and `faceEmbedder.debugFaceCrop()` → `synchronized(faceDetector)`. Meanwhile, the `auditExecutor` may still be running the next sample's embedder calls (since `capture()` returns immediately after queuing). So you can have `auditExecutor` calling `embedWithFallback()` → `segmenter.segmentAndCrop()` while `captureExecutor` calls `debugMaskedCrop()` → `segmenter.segmentAndCrop()` — both hit the `@Synchronized` monitor, so correctness is fine, but this doubles the contention window for the production pipeline when audit is on. The composite's embedder calls are redundant with the stability embedder calls — they both run on the same `(frame, box)` input.

Consider: when `saveComposite` is true, could the `auditExecutor` pass the already-computed bitmaps (or at least the snapshot) directly to `writeComposite` instead of having `captureExecutor` re-run the segmenter/BlazeFace? Or just run the composite build on `auditExecutor` itself (it's already low-priority, and the single-threaded pool serializes work anyway). This would eliminate the `captureExecutor`'s concurrent embedder calls entirely.

Not a correctness bug — just avoidable contention for the "audit on" workflow.

**2. Unused variable `tileBlockHeight`**
`CropDebugCapture.kt` line 380:
```kotlin
val tileBlockHeight = HEADER_HEIGHT + TILE_WIDTH + PADDING * 2  // worst-case square tile
```
Computed but never read — the actual heights come from `tileHeights`. Dead code.

---

### Low

**3. `AUDIT_STABILITY_INTERVAL` / `AUDIT_COMPOSITE_INTERVAL` are instance `val`s, not companion constants**
`ObjectTracker.kt` lines 703-706. These are conceptual constants (they never change). Should be in `companion object` or at least `const val`. Currently they look like per-instance tunable state, which is misleading — and `stabilityLogger.samplingIntervalFrames` is set from `AUDIT_STABILITY_INTERVAL` in two places (init + `applyPendingLockIfAny`), implying the interval could change, when it can't.

**4. `Ring` constructor parameter unused**
`EmbeddingStabilityLogger.Ring` takes `capacity` in the constructor but stores nothing — every method receives `capacity` as a parameter. The constructor parameter could be dropped or the class could store it as a field instead of threading it through every call.

---

- The "embedders are already-synchronized" comment on `auditExecutor` (ObjectTracker line 712) is partially true: `PersonReIdEmbedder.embedFromCrop()` and `FaceEmbedder.embedFaceFromCrop()` are `@Synchronized`, and `ObjectSegmenter.segmentAndCrop()` is `@Synchronized`. But `AppearanceEmbedder.embedBitmap()` has no Kotlin-level synchronization — it relies on MediaPipe's internal native task runner serialization. The PR body correctly identifies the contention (and defaults to off because of it), so the comment is just slightly imprecise.

- The `clearLock` drain via `auditExecutor.submit {}.get(2s)` is correct — single-threaded executor guarantees FIFO, so the no-op task completes only after all preceding work. 2s timeout is generous given ~50ms per audit sample.

- The stability table in `embedding_audit_baseline.md` is excellent — clear presentation of the noise floor per embedder, directly motivating the Phase 1 PRs. The "what the numbers say" section turns raw data into actionable priorities.

- The baseline doc still says "currently flaky — see 'Known issues' above" in the "How to reproduce" section (line 1015), but the "Known issues" section was removed in this commit and replaced with "Notes from baseline collection". Minor copy mismatch.

---

### Verdict

Clean instrumentation PR. Default-off eliminates any production risk. The stability data is valuable and well-documented. The two mediums are minor optimizations for the audit-on path, not correctness issues. Good to merge.

tuky191:
Review addressed in df5ac060:

**Med 1 (third concurrent caller):** Dropped `CropDebugCapture.captureExecutor` and routed composite-write work through `auditExecutor` via constructor injection. The audit path now serializes everything — stability sampling and composite writing — on one thread, so audit never adds more than one concurrent caller into the `@Synchronized` segmenter / BlazeFace monitors. The duplicate-embedder-call observation (audit's `embedWithFallback` → segmenter and composite's `debugMaskedCrop` → segmenter both fire on the same `(frame, box)`) is still there in the code, but with single-threaded serialization the contention window is bounded; deduping by passing already-computed crops through is a bigger refactor that can land with Phase 1's input fixes.

**Med 2 (unused `tileBlockHeight`):** Deleted.

**Low 3 (constants on instance):** Moved `AUDIT_STABILITY_INTERVAL` and `AUDIT_COMPOSITE_INTERVAL` to `ObjectTracker.companion` as `const val`. Also dropped the redundant `samplingIntervalFrames` assignment in `applyPendingLockIfAny` since it never changes after init.

**Low 4 (Ring constructor parameter unused):** `Ring.capacity` is now a `private val` field; `add()` and `nthBack()` no longer take it as a parameter.

**Doc nit:** Replaced the stale "currently flaky — see Known issues above" in How to reproduce with the actual current behavior (~44 min with audit on, 2 expected audit-induced failures).

239 unit tests pass; main + androidTest both compile.

The "imprecise comment" observation about which embedders are `@Synchronized` is fair — but with the single-thread audit-executor change, only one audit thread ever calls into any embedder at a time, so the synchronization story is now: production thread + at-most-one audit thread, regardless of which embedders use Kotlin `@Synchronized` vs MediaPipe's internal serialization. The `auditExecutor` doc comment now reflects that.

tuky191 (commented):
## Follow-up review — fix commit df5ac06

All four findings addressed:

1. **Medium #1 (captureExecutor contention)** ✅ — `CropDebugCapture` no longer owns a separate executor. Receives the shared `auditExecutor` via constructor injection — all audit work (stability sampling + composite writing) serializes on one thread. Comment updated.
2. **Medium #2 (unused `tileBlockHeight`)** ✅ — Removed.
3. **Low #3 (instance vals → companion constants)** ✅ — Moved to `companion object` as `private const val`. Redundant second assignment removed.
4. **Low #4 (Ring capacity threading)** ✅ — `Ring` stores capacity as a field; methods no longer take it as a parameter.

Bonus: stale "currently flaky" doc comment updated.

Good to merge.

Issue #100, opened 2026-04-27
Centralized canonical-crop preparation — single aspect-preserving crop fed to all embedders (#91 strategic)


tuky191:
**Strategic prerequisite for #93 / #94 / #95.** Validation pass on #92 baseline (PR #99) revealed that per-embedder letterbox/masking fixes don't compose well — every model currently owns its own crop+resize, each producing a *differently-distorted* view of the same subject. Fixing this centrally is cheaper than fixing it three times in three shapes that don't agree.

## What the audit revealed

Each embedder consumes a different distorted version of the bbox:

| Stage | Input dim | Source aspect handling | Resulting distortion |
|---|---|---|---|
| Bbox crop | arbitrary (e.g. 73×258, 28×50, 146×133) | — | — |
| MNV3 | 224×224 | MediaPipe internal (probably stretch) | square-stretch |
| OSNet | 256×128 | **stretch to 1:2** | severe on near-square crops |
| MobileFaceNet | 112×112 | **stretch to 1:1** + no alignment | rotation + aspect |
| Segmenter | 512×512 | **stretch to 1:1**, mask returned in stretched coords | mask predicted on warped imagery |
| BlazeFace | runs on the person crop (arbitrary aspect) | inherits person-crop distortion | keypoint accuracy degraded |

So:

- The boy at \`boy_indoor_wife_swap\` lock (\`146×272\` ≈ 1:1.85) and the boy at f=150 (\`146×133\` ≈ 1:1) get fed to OSNet as completely different shapes — same subject, two seconds apart, two stretched-differently inputs.
- The segmenter on \`boy_indoor\`'s \`73×258\` crop predicts on a 1:3.5 → 1:1 stretched 512×512 — its training distribution didn't include warps that severe, which probably explains the **Swiss-cheese masks** observed on chair / flowerpot (mask has holes inside the foreground).
- BlazeFace runs on the person crop, which is itself warped, so the keypoints it returns are in warped space; #93's alignment would have to undo both.

## What \"canonical crop\" means

One source-of-truth crop preparation that all embedders consume. Schematic:

\`\`\`kotlin
data class CanonicalCrop(
    val bitmap: Bitmap,             // letterboxed at the target aspect, neutral-fill padding
    val mask: ByteArray?,           // segmentation mask in the SAME pixel space, when available
    val keypoints: List<PointF>?,   // face keypoints in the same space, when applicable
[...]

Issue #98, opened 2026-04-27
Min-bbox-size guard at lock time (#92 follow-up)


tuky191:
Found during the #92 baseline validation pass. The audit captured a \`person_playground_tracking_rate\` lock on a **28×50-pixel** bbox of a far-away subject:

![person_playground tiny lock](https://github.com/tuky191/haptic-track/blob/master/docs/embedding_audit_examples/person_playground_lock.jpg?raw=true)

At that resolution:

- Segmenter \"succeeds\" but the mask is mostly empty pixels (latches onto a small foreground blob)
- OSNet upscales 28→128 px (4.5×) — severe pixelation
- BlazeFace finds no face (subject too small)
- MNV3 k=1 p10 = **0.20** on this video — embedder is essentially embedding noise

No Phase 1 input-quality fix (#93/#94/#95) can rescue inputs this small. The cleanest fix is **don't allow a lock on a bbox below a minimum pixel area** — refuse the lock or expand it via context if there's no acceptable bbox.

## Suggested behavior

In \`ObjectTracker.lockOnObject\`:

\`\`\`kotlin
val minPixelW = (frameW * 0.04f).toInt()  // ~4% of frame width
val minPixelH = (frameH * 0.04f).toInt()
val pixelW = (boundingBox.width() * frameW).toInt()
val pixelH = (boundingBox.height() * frameH).toInt()
if (pixelW < minPixelW || pixelH < minPixelH) {
    Log.w(TAG, \"Lock rejected: bbox too small (\${pixelW}×\${pixelH}, min=\${minPixelW}×\${minPixelH})\")
    return  // or surface to UI
}
\`\`\`

Numbers want tuning — the audit baseline shows that 28×50 px is unworkable but doesn't tell us where the cliff is. A short experiment: re-lock the same person_playground subject at 60×100, 80×140, 120×210 and look at the audit's same-vs-different cosines.

## Out of scope

Auto-zoom-to-fit before lock. That's a UX change rather than an instrumentation fix and probably wants the user's input.

## Why now (vs Phase 1)

Doesn't block Phase 1. But Phase 1 baselines are noisier than they need to be because tests like \`person_playground\` include pathological tiny-bbox locks that drag down the per-video noise floor. Filing it so the data we collect for #93/#94/#95 deltas is on cleaner inputs.

Issue #91, opened 2026-04-27
Audit + fix embedding input pipeline before swapping models (alternative to #69)


tuky191:
## Why this issue

Before committing to #69 (replace MobileNetV3 with MobileCLIP-S2 / DINOv2-small, +50–80MB APK, 3× latency), audit whether our current embedders are even seeing clean inputs. A model swap is a poor fix for noisy crops — the new model would inherit the same problems.

Audit scope: the four embedding paths and the segmenter that feeds them.

- `AppearanceEmbedder.kt` (MNV3, 1280-dim, generic appearance)
- `PersonReIdEmbedder.kt` (OSNet x1.0, 512-dim, body re-id)
- `FaceEmbedder.kt` (MobileFaceNet, 192-dim, face identity)
- `ObjectSegmenter.kt` (magic_touch, used to mask background before MNV3)

### A. Background contamination — OSNet and face skip the segmenter ⚠️ high ROI

`AppearanceEmbedder.embed()` runs `segmenter.segmentAndCrop` and embeds the masked crop (background zeroed). `PersonReIdEmbedder.embed()` and `FaceEmbedder.embedFace()` both call \`cropNormalized\` directly — raw rectangle including background.

- \`PersonReIdEmbedder.kt:69\` — \`cropNormalized(bitmap, personBox)\`
- \`FaceEmbedder.kt:74\` — \`cropNormalized(bitmap, personBox)\`
- \`ObjectTracker.kt:241,442,471,472,741,743,1019,1021\` — every OSNet/face call site uses the bbox path, never the masked crop

Asymmetry: the cleanest input goes to the embedder we lean on least (MNV3, 10% of scoring). The signals we lean on most (OSNet gate #67, face gate #84, scene memory #89) embed wallpaper, sofa, the person standing behind the lock.

**Fix:** thread the masked crop produced once at lock/embed time through to OSNet. For face, run BlazeFace on the masked person crop, then crop the face from the masked region — but verify masked-out shoulders don't suppress face detection.

### B. No face alignment ⚠️ highest single-fix ROI

\`FaceEmbedder.embedFaceFromCrop()\` (\`FaceEmbedder.kt:87\`) crops the BlazeFace bounding box and resizes it 112×112. BlazeFace produces 6 keypoints (eyes, nose, mouth, ears) — they're ignored.

[...]

PR #101, opened 2026-04-27
Canonical-crop preparation: aspect-preserving inputs across all embedders (#100)

### LINKED ISSUE #95: Phase 1C (#91): letterbox aspect ratio across all three embedders
Parent: #91 — embedding input audit.
Depends on: Phase 0 baseline. Best landed after #93 (face alignment) since alignment supersedes the face-side change.

## Problem

All three embedders use \`Bitmap.createScaledBitmap(crop, W, H, true)\` — bilinear stretch, no aspect preservation:

- OSNet target 256×128 (2:1). Sitting persons, half-body crops, low-confidence boxes give non-2:1 crops → squashed vertically or stretched horizontally
- MNV3 target 224×224. Real objects (water bottle, lamp, mug) are rarely square → distortion
- Face target 112×112. BlazeFace bboxes are roughly square; smaller concern but not zero

Same object across consecutive frames with slightly different bbox aspect ratios → different embeddings of the same object → noise in self-similarity (k=1).

## Change

Add \`padToAspectThenResize(bitmap, targetW, targetH, fillColor)\` to \`EmbeddingUtils.kt\`:

- Pad source to target aspect ratio with neutral fill (mean pixel or black — verify on device which is better via Phase 0 composites)
- Then resize to target dimensions
- Apply at:
  - \`PersonReIdEmbedder.embedFromCrop\` (\`PersonReIdEmbedder.kt:85\`) — 256×128
  - \`FaceEmbedder.embedFaceFromCrop\` (\`FaceEmbedder.kt:105\`) — 112×112; **no-op if #93 (alignment) lands first** since alignment already produces canonical 112×112
  - MNV3 path — 224×224; verify what MediaPipe \`ImageEmbedder\` does internally first. If it stretches, wrap our crop. If it letterboxes already, no-op.

## Measurement (vs Phase 0 baseline)

- Primary signal: k=1 self-similarity should **rise** (less aspect-ratio jitter between consecutive frames of the same object)
- Wrong-identity reacqs: secondary signal, may or may not move
- Same-instance vs different-instance separation: should widen modestly

## Risks

Small. Worst case is no-op for one or more embedders.

## Code refs

- \`EmbeddingUtils.kt\` — where \`padToAspectThenResize\` lands
- \`PersonReIdEmbedder.kt:85\` — \`Bitmap.createScaledBitmap\` call
- \`FaceEmbedder.kt:105\` — same
- \`AppearanceEmbedder.kt:38-44\` — MediaPipe \`ImageEmbedder\` config; check whether internal resize is letterbox or stretch

tuky191:
## Summary
- Strategic prerequisite for #93 / #94 (closes #95 / #98 as superseded). Per-embedder letterbox fixes don't compose because each model owns its own crop+resize and produces a differently-distorted view of the same subject (#91 audit findings).
- Introduces `CanonicalCrop` + `CanonicalCropper` as the source-of-truth preparation feeding MNV3, OSNet, MobileFaceNet, and the magic_touch segmenter. Replaces the per-embedder `cropNormalized` + stretch-resize idiom with a uniform aspect-preserving letterbox at each model's native input dim.
- Public API of all four embedders preserved — `ObjectTracker` is unchanged. The new canonical-input methods sit alongside the existing `(bitmap, RectF)` overloads.

## Why now (verified evidence)
**Production aspect distribution** from 16 audit sessions (n=11,082 detections):
- Person bboxes: median aspect 0.967 (near-square), only **9% within OSNet's 1:2 native target (0.4–0.6)**, **69% near-square or wider** — currently being stretched horizontally by 2× to fit OSNet's 256×128.
- Other labels (chair, oven, keyboard, etc.) similarly diverse.

**MediaPipe ImageEmbedder default**: `keep_aspect_ratio = false` (verified in the proto) — stretches when input ≠ model size. Java API doesn't expose the flag, so pre-letterboxing to exact 224×224 is the only way to bypass it.

PR #99 also documented Swiss-cheese masks on heavily-warped 512×512 segmenter inputs (chair-seat-only, holes inside chair body).

## Defaults agreed in #100
- Neutral-gray (0x7F7F7F) letterbox fill — channel-balanced, less mean shift than black for ImageNet-trained models. Empirical validation deferred to the audit re-baseline.
- 5% bbox padding uniform across targets.
- 28×28 minimum source pixels — folds in the spirit of #98.
- Caller of `CanonicalCropper.prepare` owns the canonical bitmap and recycles after consumers are done.

## Files
| File | Change |
|---|---|
| `CanonicalCrop.kt` (new) | Types + cropper. ~140 LoC. |
| `CanonicalCropperTest.kt` (new) | 17 tests: aspect preservation, padding correctness, min-pixel guard, edge clamping, source-rect round-trip. |
| `AppearanceEmbedder.kt` | New `embed(canonical)`. Old API preserved as wrappers; segmentation now runs on a 512² canonical and downsamples masked output to 224² for MNV3. |
| `PersonReIdEmbedder.kt` | New `embed(canonical)`. OSNet sees letterboxed 128×256 input. |
| `FaceEmbedder.kt` | New `embedFace(personCanonical)`. Person canonical 256×256; BlazeFace runs on canonical; face sub-canonical 112×112 (no 5-point alignment yet — that's #93, keypoints surfaced via `debugFaceCrop`). |
| `ObjectSegmenter.kt` | New `segmentCanonical(canonical)` runs on 512×512 letterboxed input; mask in canonical space. Legacy `segmentAndCrop` and `extractContour` are wrappers. |

Diff: 6 files changed, 718 insertions, 227 deletions.

## What's NOT in this PR
- **#93 face alignment** — 5-point similarity transform on canonical keypoints. Lands on top of this. Keypoints already exposed in canonical space via `FaceEmbedder.debugFaceCrop` so #93 can wire up alignment without further refactor.
- **#94 mask OSNet input + structural mask quality gating** — Swiss-cheese detection (connected-component count, top-component area share). Will consume `segmentCanonical`'s canonical-space mask.
- Threading the canonical-space mask into OSNet — also part of #94.

## Test plan
- [x] 256 unit tests pass (239 existing + 17 new), 0 failures.
- [x] `assembleDebug` succeeds.
- [x] `assembleDebugAndroidTest` succeeds.
- [ ] **Audit re-baseline pending** (this is why the PR is draft):
  - Set `CropDebugCapture.AUDIT_ENABLED = true`, build + install on device.
  - Verify `test_videos/` sizes match local before launching (per `feedback_video_file_drift` — stale 4K file looks like a hang).
  - Run full instrumented suite (~44 min). Expected 2 audit-induced contention failures (`man_desk_camera_swing_tracking_rate`, `mouse_desk_rotation_reacquires_correctly`) — those reflect contention, not regressions.
  - Pull `debug_frames/`, regenerate `docs/embedding_audit_baseline.md`, visual-diff a few composites:
    - **Hold the line**: man_desk k=1 p10 ≥ master baseline.
    - **Expected gain**: tests with non-2:1 person bboxes (boy_indoor, two_men_*) — OSNet k=1 should rise as aspect distortion drops.
  - Flip `AUDIT_ENABLED = false`, commit re-baseline doc, mark ready-for-review.
- [ ] Manual smoke test on device: lock + reacquire still work end-to-end after the refactor.

## Closes / supersedes
- Supersedes #95 (per-embedder letterbox absorbed here).
- Folds in #98 (min-bbox-size guard).
- Unblocks #93, #94.
- Lays groundwork for the Phase 1 → #69 decision gate.

### DIFF
--- a/app/src/main/java/com/haptictrack/tracking/AppearanceEmbedder.kt
+++ b/app/src/main/java/com/haptictrack/tracking/AppearanceEmbedder.kt
@@ -18,16 +18,30 @@ import com.google.mediapipe.tasks.vision.imageembedder.ImageEmbedder
  * similarity against the stored reference.
  *
  * Model: MobileNetV3 Large — ~10MB, ~8ms per crop.
+ *
+ * #100 input flow: every embed call consumes a [CanonicalCrop] at exactly
+ * [MNV3_INPUT_SIZE]² so MediaPipe's internal `keep_aspect_ratio=false`
+ * stretch is bypassed — the model sees uniformly aspect-preserved input
+ * across every detection. When segmentation succeeds, the masked canonical
+ * is fed instead of the raw canonical so the model embeds foreground only.
  */
-class AppearanceEmbedder(context: Context) {
+class AppearanceEmbedder(
+    context: Context,
+    private val cropper: CanonicalCropper = CanonicalCropper(),
+) {
 
     companion object {
         private const val TAG = "AppearEmbed"
         private const val MODEL_PATH = "mobilenet_v3_large_embedder.tflite"
+        /**
+         * MobileNetV3 Large native input. We pre-letterbox to this so MediaPipe
+         * skips its internal resize and can't silently re-stretch our inputs.
+         */
+        const val MNV3_INPUT_SIZE = 224
     }
 
     private val embedder: ImageEmbedder
-    private val segmenter: ObjectSegmenter = ObjectSegmenter(context)
+    private val segmenter: ObjectSegmenter = ObjectSegmenter(context, cropper)
 
     init {
         embedder = try {
@@ -56,6 +70,18 @@ class AppearanceEmbedder(context: Context) {
     )
 
     /**
+     * Compute an MNV3 embedding from a prepared [CanonicalCrop]. Caller owns
+     * [canonical] and is responsible for recycling. Required dims: 224×224.
+     */
+    fun embed(canonical: CanonicalCrop): FloatArray? {
+        if (canonical.targetWidth != MNV3_INPUT_SIZE || canonical.targetHeight != MNV3_INPUT_SIZE) {
+            Log.w(TAG, "Canonical dims ${canonical.targetWidth}×${canonical.targetHeight} != expected ${MNV3_INPUT_SIZE}²")
+            return null
+        }
+        return embedBitmap(canonical.bitmap)
+    }
+
+    /**
      * Extract an embedding using segmentation masking. Returns null if segmentation
      * fails — caller should NOT match against unmasked embeddings since they include
      * too much background noise for reliable identity discrimination.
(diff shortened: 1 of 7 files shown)


tuky191 (commented):
## Code Review — PR #101 (Canonical-crop preparation, #100)

**Scope:** Introduces \`CanonicalCrop\` + \`CanonicalCropper\` as the uniform aspect-preserving letterbox preparation for all 4 embedders (MNV3, OSNet, MobileFaceNet, magic_touch segmenter). Replaces per-embedder \`cropNormalized\` + stretch-resize with letterbox at each model's native input dim. Old API preserved as wrappers. 718 insertions, 227 deletions, 17 new tests.

275 unit tests pass.

---

### Medium

**1. \`masked\` bitmap leaks in \`embedWithAugmentations\` on OOM**
\`AppearanceEmbedder.kt\` line 180: if \`Bitmap.createScaledBitmap(masked, MNV3_INPUT_SIZE, MNV3_INPUT_SIZE, true)\` throws (OOM), the \`masked\` bitmap allocated at line 171-174 is never recycled — the exception propagates before reaching line 210 which returns \`AugmentedResult(results, masked)\` (caller recycles). Low real-world risk (OOM is already a crisis), but the fix is a simple \`try/catch\` around the \`val (mnv3Source, ownsSource)\` destructuring that recycles \`masked\` on failure.

**2. \`segmentAndCrop\` legacy wrapper: comment says "Crop after mask" but the Bitmap.createBitmap sub-region call creates a copy**
\`ObjectSegmenter.kt\` line 158: \`Bitmap.createBitmap(masked, pad.left, pad.top, drawW, drawH)\` followed by \`masked.recycle()\` in the finally block. This is CORRECT — Android's sub-region createBitmap copies pixel data when the region is a proper subset — but the pattern is subtle. A one-line comment (\`// createBitmap copies pixels, so masked recycle is safe\`) would prevent future contributors from questioning it.

**3. Contour back-mapping clamps letterbox points to source edge**
\`ObjectSegmenter.kt\` \`extractContour\`: contour points that land in the gray letterbox area are clamped to \`[0,1]\` via \`coerceIn\`, mapping them to the crop edge rather than discarding them. This means contours near the subject boundary may include artifact points from the letterbox region. Minor accuracy issue — contours are already approximate (\`approxPolyDP\`).

---

### Low

**4. \`CanonicalCropper\` constructs a new instance per call**
\`cropper.prepare()\` is called multiple times per frame (MNV3 canonical, segmenter canonical, OSNet canonical). The cropper itself is stateless — the allocation cost is negligible. Just noting there's no pooling if it ever became expensive.

**5. Test coverage gap: no test for non-square source bitmaps**
The 17 CanonicalCropperTest tests cover square/landscape/portrait targets and padding, but all source bitmaps are 200×100 or similar synthetic sizes. No test exercises a real-world 1920×1080 frame → 128×256 OSNet canonical path. The math is the same, but a coverage test would catch any integer-overflow edge cases at large dimensions.

---

### Observations

- The letterbox math in \`CanonicalCropper.prepare()\` is correct: \`fitScale = min(targetW/drawW, targetH/drawH)\`, padding computed from the remainder. The 5% bbox padding + min-pixel guard (28×28) folds in #98 cleanly.
- MNV3 now sees the segmenter mask downscaled to 224² instead of stretched — this should directly improve the MNV3 noise floor numbers from the audit baseline.
- OSNet's 128×256 letterbox input preserves aspect ratio that was previously destroyed by the 2× vertical stretch. The audit table predicted "tests with non-2:1 person bboxes should see OSNet k=1 rise" — this PR delivers that.
- The old API wrappers (\`embed(bitmap, box)\`, \`segmentAndCrop(bitmap, box)\`) correctly delegate to the canonical path, so \`ObjectTracker\` is unchanged.

---

### Verdict

Solid foundation PR. The letterbox math is correct and the refactor is clean. The one medium (masked bitmap leak on OOM) is low real-world risk. Ready for on-device audit re-baseline (the stated gate for converting from draft).

tuky191 (commented):
## Follow-up review — PR #101, commits 2f734807 + 6b393217

Two findings addressed:

1. **Medium (masked bitmap OOM leak)** ✅ — `embedWithAugmentations` now wraps `createScaledBitmap` in try/catch, recycles `masked`, and returns `AugmentedResult(emptyList(), null)` on failure. Clean.
2. **Medium (segmentAndCrop sub-region recycle clarity)** ✅ — Comment added explaining Android's copy-on-sub-region API contract.

Also picks up #111 (tap-to-track auto-record) as expected from the stack.

278 unit tests pass. Good to merge (pending audit re-baseline).

PR #111, opened 2026-04-28
Tap-to-track auto-starts recording

tuky191:
## Summary

When the user taps on a detection to lock onto it, automatically start recording. Removes the second tap currently needed to start capture after locking.

One-line change in `CameraViewModel.onTapToLock`: after `lockOnObject` succeeds, if `!isRecording` toggle recording on.

## Why

Hands-free flow: tap → lock + record. The volume-down hands-free cycle (idle → lock → record) already works this way; tap-to-track was the only path that required two interactions. This makes them consistent.

## Test plan

- [ ] Tap to track an object on device → lock indicator appears AND recording starts
- [ ] Volume-down cycle still works (idle → lock+record → stop+clear)
- [ ] Tap an already-recording session re-locks without disrupting the recording
- [ ] No double-recording or lost frames at the lock moment

### DIFF
--- a/app/src/main/java/com/haptictrack/ui/CameraViewModel.kt
+++ b/app/src/main/java/com/haptictrack/ui/CameraViewModel.kt
@@ -163,6 +163,7 @@ class CameraViewModel(application: Application) : AndroidViewModel(application) 
         if (tapped != null) {
             objectTracker.lockOnObject(tapped.id, tapped.boundingBox, tapped.label)
             _uiState.update { it.copy(status = TrackingStatus.LOCKED, trackedObject = tapped) }
+            if (!_uiState.value.isRecording) toggleRecording()
         }
     }
 


tuky191 (commented):
## Code Review — PR #111 (Tap-to-track auto-starts recording)

One-line change in `CameraViewModel.onTapToLock`: after `lockOnObject` succeeds, toggle recording on if not already recording.

275 unit tests pass.

---

### Low

**1. Race with lock failure**
`lockOnObject` dispatches work to `lockExecutor` and returns immediately (off-thread lock, PR #66). The `toggleRecording()` call fires before the lock result is applied — if the lock fails (e.g., no bitmap available), recording starts without a lock. In practice this is a narrow window and the user can stop recording, but worth noting.

---

### Verdict

Trivially correct for the happy path. Aligns tap-to-track with the volume-down cycle. Good to merge — verify the timing on device (the off-thread lock might cause a brief recording-without-lock flash).

Issue #117, opened 2026-04-29
OSNet-IBN MSMT17 TFLite conversion produces degenerate output


tuky191:
## What

The osnet_ibn_x1_0_msmt17.tflite shipped with #113 produces near-constant output on-device:

\`\`\`
PersonReId: OSNet out=[0.0000, 0.0322, 0.0000, 0.0000, ...]
PersonReId: OSNet out=[0.0000, 0.0322, 0.0000, 0.0000, ...]
PersonReId: OSNet out=[0.0000, 0.0322, 0.0000, 0.0000, ...]
PersonReId: OSNet out=[0.0000, 0.0528, 0.0000, 0.0000, ...]   ← rare outlier
PersonReId: OSNet out=[0.0000, 0.0322, 0.0000, 0.0000, ...]
\`\`\`

Across 100+ inferences with different person crops, the output is essentially the same vector — \`cosineSimilarity\` between any two outputs ≈ 1.000000.

The old osnet_x1_0_market.tflite produces correctly varied output:
\`\`\`
PersonReId: OSNet out=[0.0039, 0.0000, 0.0358, 0.0000, ...]
PersonReId: OSNet out=[0.0000, 0.0000, 0.0851, 0.0000, ...]
PersonReId: OSNet out=[0.0446, 0.0257, 0.0375, 0.0000, ...]
\`\`\`

## Impact (live)

The degenerate output caused all reId scores to saturate to 1.000 in the cascade, leading to wrong-person reacquires across multiple sessions (see #116). #116 reverts to osnet_x1_0_market.tflite as the unblocking fix.

## What needs investigating

Conversion recipe is in \`tools/models/CONVERT_OSNET_IBN.md\` (Docker for onnx2tf). Round-trip cosine-vs-PyTorch was claimed at 0.99996 in PR #113's commit message. Either:

1. The PyTorch reference itself was off (reference run not actually meaningful)
2. The TFLite output is correct on CPU but the TFLite GPU delegate breaks IBN-specific layers
3. Some preprocessing mismatch (the input we're feeding doesn't match what the converted graph expects)

## Reproduction

1. Apply osnet_ibn_x1_0_msmt17.tflite as MODEL_ASSET
2. Add log line in PersonReIdEmbedder.embed() printing first 4 floats of \`embedding\` after l2Normalize
3. Run any VideoReplayTest that exercises persons (e.g. \`walking_outdoor_diagnostic\`)
4. Observe near-constant output

## Next step

[...]

Issue #116, opened 2026-04-29
Walking session: 12 reacqs in 14s, REACQUIRE frames saved upside-down


tuky191:
## What happened

Live device session 2026-04-29 09:41 (\`session_20260429_094142_person\`, 433 frames) — outdoor walking, locked on a person wearing dark/blue clothing, then walked while filming. The lock dropped and reacquired **12 times** in ~14 seconds.

Pulled the session to \`pulled_frames/session_20260429_094142_person/\`.

### 1. Cascade scoring admits clearly-impostor candidates
\`reId=1.000\` dominates the cascade even when MNV3 strongly disagrees:

\`\`\`
REACQUIRE id=327 sim=0.193 reId=1.000 znMnv3=-2.236   ← znMnv3 z-score -2.2 says impostor; still admitted
REACQUIRE id=345 sim=0.616 reId=1.000 znMnv3=-0.397
REACQUIRE id=399 sim=0.426 reId=1.000 znMnv3=0.388
REACQUIRE id=497 sim=0.306 reId=1.000 znMnv3=0.119
REACQUIRE id=510 sim=0.383 reId=1.000 znMnv3=0.612
\`\`\`

Same root cause as #102's kid_to_wife but in a single-person scene where SessionRoster open-set rejection cannot fire (no non-lock slots observed). 0 \`ROSTER_REJECT\` and 0 \`FACE_GATE_REJECT\` events across the session.

This is Phase 4 of #102 (scoring-level z-norm). Tracking it here as the real-world reproducer.

### 2. Saved REACQUIRE frames are upside-down

Multiple \`*_REACQUIRE_raw.png\` files in the session show the scene rotated 180° from the LOCK frame (sky on bottom, ground on top). It's unclear whether:

- The engine processed upside-down bitmaps (production bug — would explain \`sim=0.19\` and \`znMnv3=-2.2\` directly), OR
- The engine processed upright bitmaps but \`DebugFrameCapture\` saved the unrotated source (cosmetic).

Reproducing this on-device is the goal of the planned diagnostic test (this issue's first action item).

## What we'll do

1. Pull the source 4K video for this session and one of the supermarket sessions (\`session_20260429_101614_person\`) into \`test_videos/\`
2. Add diagnostic VideoReplayTest cases — log everything, no hard assertions
3. Re-run on device and compare:
   - Saved frames during replay vs during live session
   - znMnv3 distribution and \`reId=1.000\` frequency
[...]

PR #113, opened 2026-04-28
SessionRoster + OSNet-IBN MSMT17 (Phase 4 of #102, closes #108)

tuky191:
## Summary

Closes #108. Closes the kid_to_wife admission path that #102 phase 3 (z-norm calibration) couldn't fix. Two independent changes shipped together as the "all features" test build:

### 1. SessionRoster — structural identity-by-rank (commit \`99a8fcbc\`)

Field consensus from MOT (DeepSORT, StrongSORT, BoT-SORT) and SOT (DaSiamRPN, DAM4SAM): identity is a RANK problem with a margin to second-best, not an absolute threshold to a single anchor. The cascade was asking "is candidate similar to lock?" but never "is candidate more similar to lock than to other people we've seen?"

\`SessionRoster.kt\` adds a clustered tracklet store keyed by face/body cosine. Slot 0 = lock; slots 1..N = other persons seen in the session. At reacquire time, a candidate's identity is judged by RANK among all known identities + a margin to the second-best — not by an absolute threshold to the lock alone.

The existing \`SCENE_FACE_VETO\` + \`SCENE_BODY_VETO\` (#83 phase 2) is subsumed by a single \`ROSTER_REJECT\` gate with uniform handling across modalities. Aggressive observation: every other person at lock time + every 5 frames during VT-confirmed tracking, with partial face-only or body-only observations accepted.

19 new unit tests (\`SessionRosterTest\`) covering seed/observe/cluster/merge/evict + a structural kid_to_wife regression.

### 2. OSNet-IBN x1.0 MSMT17 swap (commits \`9ca6c7a6\`, \`35370c6c\`, \`c76026be\`, \`f581be12\`)

Off-device benchmark on real same-vs-different person crops (extracted from kid_to_wife, boy_indoor_wife_swap, man_desk via YOLOv8+ByteTrack):

| Model | kid_to_wife gap | boy_swap gap |
|---|---:|---:|
| OSNet x1.0 Market (current TFLite) | +0.071 | -0.226 |
| OSNet x1.0 MSMT17 | +0.029 | -0.225 |
| OSNet-AIN x1.0 MSMT17 | +0.038 | -0.177 |
| **OSNet-IBN x1.0 MSMT17** | +0.005 | **-0.162** |

(gap = same_p10 − diff_p90; positive = fixed threshold separates same vs different)

[...]

Issue #102, opened 2026-04-27
Wrong-identity reacquires: re-id noise floor overlaps the override threshold (structural fix needed)


tuky191:
## Summary

Re-acquisition fires onto **clearly different people** at OSNet cosine similarities the engine treats as same-identity (0.7+ override / 0.45+ minimum). This is a structural property of the embedders on this device against this domain, not a bug in any single component, and the literature is explicit about both the cause and the fixes.

This is currently the most impactful failure mode the tracker has — the entire identity-aware re-acquisition pipeline can be defeated by a brief lock + camera pan to any other indoor person.

## Concrete evidence (live device session, 2026-04-27)

Lock on small distant kid in doorway → 2s later, camera panned and **wife reacquired with `OSNet=0.905` and engine-reported `MNV3 sim=0.864`** (above the 0.7 embedding-override threshold).

What forensic instrumentation revealed (`session.log` augmented with `ACCUM`/`TEMPLATE`/`DRIFT` lines):
- Recomputing MNV3 from `scenario.json`'s lock gallery alone gives **best-gallery sim = 0.541** vs the wife — clean different-person territory.
- The engine's logged `0.864` came from the live gallery, which had grown by one entry during the 2s VT window. The accumulator's gate (`centroidSim < 0.92 OR gallery.size < 8`) waved in non-lock embeddings during VT drift (`lockSim ≈ 0.06`–`0.08`), poisoning the gallery. _A separate small PR is being prepared for this gate fix; it's necessary but **not sufficient**._
- Synthetic on-device OSNet experiment with neutral inputs (`OsnetGrayBiasTest`):
  - Different-person noise floor on this device: **0.620 native-aspect, 0.667 stretched, 0.686 letterboxed** (with gray fill).
  - Pure-gray vs letterboxed-person canonical: **0.671** — a non-trivial fraction of OSNet's response is "gray response", overlapping identity space.
  - Same-source under different geometric treatments: **0.87–0.90** — identity preserved, but the gap to different-person is only **~0.18**.
- The wife's `reId=0.823` is _consistent_ with the device noise floor + content overlap; not a single-component bug.

## Documented in literature

Cross-domain Rank-1 of Market-1501-trained models drops 30–60 points on out-of-domain footage; mAP drops 40+ points. Indoor handheld + kid-scale is decisively off OSNet's training manifold. Selected references:

[...]

PR #118, opened 2026-04-29
Diagnose + fix wrong-person reacquires: OSNet GPU miscompile + cache pollution (#116, #117)

tuky191:
## Summary

The 2026-04-29 live device sessions (walking outdoor + supermarket checkout) showed wrong-person reacquires across the board with `reId=1.000` dominating the cascade. This PR diagnoses two compounding root causes and ships fixes for both.

**Headline:** the OSNet-IBN MSMT17 model bundled in #113 produces a near-constant output on Adreno GPU delegate. That made every candidate look like the lock to the cascade. Reverting to a working model + dropping a separate cache-pollution path cleans up the cascade. Adds a self-test at init that will catch any future silent miscompilation.

## Two distinct bugs

**1. OSNet-IBN MSMT17 silently miscompiled by Adreno GPU delegate (#117)**

The same TFLite file produces:

| Inference path | First 4 floats sample |
|---|---|
| Python TFLite CPU (laptop) | varied per crop, e.g. `[+0.0000, +0.0347, +0.0000, +0.0141, ...]` |
| Android TFLite CPU | varied per crop |
| Android TFLite GPU (Adreno 740) | constant `[0.0000, 0.0322, 0.0000, 0.0000, ...]` |

`cosineSimilarity` between any two outputs ≈ 1.000000 → cascade always read `reId=1.000` → wrong-person reacquires admitted on `sim=0.15` and `znMnv3=-2.2` candidates. Existing reading: TFLite GPU delegate has documented issues with InstanceNorm-decomposed-as-reduce+broadcast paths (see `tensorflow/tensorflow#51800, #27240, #56983, #46462`). Conversion math is correct (Python CPU validates); the bug is in the GPU delegate compilation step.

**2. IoU-fallback embedding inheritance** (`ObjectTracker.kt:785-796`, separate failure mode that survived even when OSNet was working)

The async cache merge previously fell back to spatial `iou > 0.3` when an ID lookup missed. After a LOST event, a new track ID at IoU>0.3 with the lock's stale `cachedBox` would inherit the lock's `reIdEmbedding` — `cosineSimilarity(lockedReIdEmbedding, candidate.reIdEmbedding)` then = 1.0 by construction. Drops to ID-only match; sync-fallback below recovers the missing-embedding case.

## What landed

| Commit | What |
|---|---|
| `cb79237a` | Diagnostic `VideoReplayTest`s for `walking_outdoor` + `supermarket_checkout` (no hard assertions, log-only) |
| `79c4a553` | Switch supermarket spec to longer 101547 source (3s slo-mo → 20.7s usable) |
| `960f1f6c` | Drop IoU-fallback embedding inheritance + revert to `osnet_x1_0_market.tflite` |
| `a412623d` | Remove broken IBN asset; add `tools/test_osnet_tflite_cpu.py` (Python validation) |
| `70ecfbe5` | OSNet model self-test on init: 3 distinct-scene crops, threshold 0.99 |
| `83c5609b` | Swap to `osnet_x1_0_msmt17.tflite` (base x1_0, MSMT17 weights, GPU-safe) |
| `887349d8` | Doc updates (`CONVERT_OSNET.md` documents both shipped + investigated-but-not-shipped variants) |

## Self-test verifies both ways

```
osnet_x1_0_market    (working): max pairwise = 0.68     → PASS
osnet_x1_0_msmt17    (shipped): max pairwise = 0.561    → PASS
osnet_ibn_x1_0_msmt17 (broken): max pairwise = 1.0000001 → FAIL → embedder auto-disabled
```

If a future model regresses the same way, `embed()` returns null and the cascade falls back to MNV3-only scoring instead of poisoning identity matching. Adds 9 KB to APK (3 small JPEG crops in assets/).

## Off-device discrimination comparison (kid_to_wife video)

| Model | same_p10 | diff_p90 | gap (same_p10 - diff_p90) |
|---|---|---|---|
| osnet_x1_0_msmt17 (shipped) | 0.573 | 0.544 | **+0.029** |
| osnet_x1_0_market (prior) | 0.717 | 0.646 | **+0.071** |

MSMT17 produces lower absolute cosines across the board (both same- and diff-person), so the gap is similar but shifted down. Cascade behavior is more conservative: in the supermarket replay, the new model produces 0 reacqs vs market's 6 — preferring "stay in LOST" over admitting impostors. Same Adreno-GPU latency (identical architecture).

## Diagnostic replay results

| Run | walking_outdoor track / reacqs | supermarket_checkout track / reacqs |
|---|---|---|
| Original (broken IBN, no fixes) | 32% / 7 | 65% / 4 (false reacqs) |
| `osnet_x1_0_msmt17` + IoU-fallback removed + self-test | 39% / 6 | 8% / **0** |

## Test plan

- [x] `./gradlew testDebugUnitTest` — all unit tests pass
- [x] `./gradlew assembleDebug assembleDebugAndroidTest` — clean build
- [x] On-device launch — self-test passes for shipped model: `OSNet self-test passed: max pairwise cosine = 0.560896 (< 0.99)`
- [x] On-device launch with broken IBN re-installed — self-test fails as expected, embedder auto-disables: `OSNet self-test FAILED: max pairwise cosine = 1.0000001 > 0.99`
- [x] `walking_outdoor_diagnostic` + `supermarket_checkout_diagnostic` — both run end-to-end, log results
- [x] `tools/.venv/bin/python tools/test_osnet_tflite_cpu.py` — Python CPU validation across all 3 model variants

## What this PR does NOT close

Both #116 and #117 stay open with reduced scope:

- **#116** — orientation half (mid-recording phone rotation produces upside-down crops fed to embedders). Pipeline-wide rotation fix (apply `runDetector`'s upright transform to VT/embedders/DebugCapture) is the structural fix; not in this PR.
- **#102 / Phase 4** — cascade still admits some impostors because `reId` carries ~60% scoring weight even at modest discrimination. Phase 4 z-norm in scoring formula is the next planned work.
- **#117** — IBN MSMT17 conversion preserved at `tools/models/tflite/osnet_ibn_x1_0_msmt17/` for future re-validation if upstream TFLite GPU delegate ever fixes the IBN miscompile. Self-test will catch any regression at init.


tuky191 (commented):
## Code Review — PR #118 (OSNet GPU miscompile + cache pollution fix, #116/#117)

**Scope:** Diagnoses and fixes two compounding root causes of wrong-person reacquires observed on device: (1) OSNet-IBN MSMT17 silently miscompiled by Adreno GPU delegate producing constant output (reId=1.000 for every candidate), (2) IoU-fallback embedding inheritance where a new detection at IoU>0.3 with the lock's stale cachedBox inherited the lock's reIdEmbedding verbatim.

295 unit tests pass.

---

### Medium

**1. Self-test runs on the GPU delegate init path — adds ~100ms to app startup**
The self-test embeds 3 JPEG crops during `PersonReIdEmbedder.init`, which runs inside `loadModelsOnce()` on the main loading path. Three inference calls at ~30ms each on GPU adds ~100ms to the loading spinner. This is a one-time cost and the safety benefit is clear, but worth documenting in the startup timing budget. If startup latency becomes a concern, the self-test could move to a background coroutine with the embedder disabled until it passes.

**2. `test_osnet_tflite_cpu.py` references `osnet_x1_0_market.tflite` which is deleted in this PR**
Line 723: `"osnet_x1_0_market (working baseline)": REPO / "app/src/main/assets/osnet_x1_0_market.tflite"`. The asset is removed (line 52-55 of the diff). The script will print `MISSING` for that entry, which is handled gracefully (line 789-790), but it's misleading to have a "working baseline" that no longer exists. Either update the path to point to the `osnet_x1_0_msmt17.tflite` as the new baseline, or remove the entry.

---

### Low

**3. Self-test uses `interpreter.run` directly, bypassing `@Synchronized` `embed(canonical)`**
`embedRawBitmap()` (line 251) calls `fillInputBuffer` + `interpreter.run` directly — it's not wrapped in `@Synchronized`. The init block runs before any tracking thread starts so there's no concurrency risk at init time, but if `selfTestMaxPairwiseCosine` were ever called outside of init, it would race with production `embed()` calls on the shared interpreter/buffers.

**4. Diagnostic VideoReplayTests have no assertions**
`walking_outdoor_diagnostic` and `supermarket_checkout_diagnostic` are log-only. Same pattern as `kid_to_wife_panning_diagnostic` from #105. Worth tightening once the baseline stabilizes.

---

### Observations

- The IoU-fallback removal is the right fix. The comment (lines 87-99) clearly explains the failure mode: new detection at IoU>0.3 with the lock's stale position inherits the lock's own reIdEmbedding → reId=1.000 by construction → cascade admits impostor. ID-only matching is identity-safe (FTFTracker IDs track the same physical object). The sync-fallback path below still recovers for the closest same-category candidate.

- The self-test design is smart — real crops instead of synthetic inputs, because the IBN failure mode is input-distribution-dependent (synthetic stripes pass but real photos collapse). Three distinct-scene crops give 3 pairwise comparisons for robustness.

- Model asset management is clean: broken IBN deleted from assets, working `osnet_x1_0_msmt17` added, both FP16+FP32 TFLite files preserved under `tools/models/tflite/` for reference.

- The `CONVERT_OSNET.md` unification is well-structured — documents both the shipped variant (section A) and the broken-but-preserved IBN (section B) with clear "NOT shipped" warnings.

- The `convert_in_docker.sh` parameterization (accept model name as argument) is a nice improvement over the hardcoded IBN-only version.

---

### Verdict

Critical fix for a production-breaking bug. The IoU-fallback removal closes the cache pollution path, the model revert eliminates the GPU miscompile, and the self-test prevents future regressions. Good to merge.

tuky191:
Addressed review feedback in `2814baf9`:

**Medium**:
- Updated `test_osnet_tflite_cpu.py` MODELS dict — `osnet_x1_0_market.tflite` removed in this PR, baseline now points at the shipped `app/src/main/assets/osnet_x1_0_msmt17.tflite`
- Documented the self-test's ~100ms init cost in the source comment, with a noted fallback path (move to background coroutine + default-disabled-until-test-passes) if startup latency tightens further. Kept synchronous since 100ms folds into the existing model-load spinner

**Low**:
- Added `@Synchronized` to `embedRawBitmap` — init-time only today but cheap insurance against future concurrent misuse of the shared `inputBuffer`/`outputArray`
- Diagnostic-tests-have-no-assertions left as-is, matching the existing `kid_to_wife_panning_diagnostic` pattern from #105; will tighten once baselines stabilize

Ready for re-review or merge.

tuky191 (commented):
## Follow-up review — PR #118, commit 2814baf9

All three findings addressed:

1. **Medium (self-test startup cost)** ✅ — Comment documents the ~100ms cost, explains it's folded into the model-load spinner, and notes the background-coroutine escape hatch.
2. **Medium (stale market baseline in Python script)** ✅ — Removed deleted `osnet_x1_0_market` entry. New baseline is `osnet_x1_0_msmt17 fp16 (shipped baseline)` at the correct asset path.
3. **Low (embedRawBitmap not synchronized)** ✅ — `@Synchronized` added with comment explaining the shared buffer hazard.

Good to merge.

Issue #110, opened 2026-04-28
chair_living_room tracking rate regressed 76% → 25% (non-person, unrelated to #108)


tuky191:
## Symptom

`chair_living_room_tracking_rate` test (in `VideoReplayTest`) fails on `feat/session-roster-108`:

```
Replay complete: 685 processed / 1373 total (687 dropped, 50% drop rate),
25% tracked, 5 reacqs, 6 losses, timedOut=false
```

Baseline (per the test's own comment, dated 2026-04-26): 76-77% tracked, 5-6 reacqs. The reacq count is unchanged but tracking rate dropped to ~⅓ of baseline.

The companion test `chair_living_room_reacquires_correctly` still passes — same number of reacquires, no wrong-category, no timeout. It's specifically the *frames-tracked* metric that regressed.

## What this means

Same number of LOST→REACQUIRE cycles, but each cycle is now spending many more frames in the "lost, searching" state before reacquiring. The chair takes longer to be re-locked after each loss.

## What is NOT the cause

- **SessionRoster (#108)**: open-set rejection is gated on `lockedIsPerson && candidateIsPerson`. For a chair lock, this entire code path is skipped. The roster does not affect non-person tracking.
- **OSNet-IBN swap (#106)**: `PersonReIdEmbedder` is invoked only for person candidates. Chair candidates use the MNV3 generic embedder.

Confirmed by re-running `chair_living_room_tracking_rate` on this branch with `cmd deviceidle whitelist` set (rules out the Xiaomi freeze that affected boy_indoor — see `feedback_xiaomi_deviceidle_freeze`). 25% reproduces.

## Likely culprit (parent stack)

The branch is stacked on multiple parent branches that touch the generic-embedder path. Candidates for bisection:

- **`1089ce27` canonical-crop preparation (#100)** — changed how MNV3 receives its crop input (aspect-preserving letterbox via `CanonicalCropper`). MNV3 is the dominant signal for non-person re-id; a preprocessing change could shift its same/different cosine bands.
- **`5929b8b0` gallery accumulator gate (#103)** — changed when MNV3 embeddings get added to `_embeddingGallery` during VT-confirmed tracking (`centroidSim < 0.92 OR gallery.size < 8` → `AND lockSim > 0.5`). For chair tracking with weaker centroid signals, the new gate may admit fewer per-angle augmentations into the gallery, narrowing identity coverage during reacquire.
[...]

PR #119, opened 2026-04-29
Phase 4 (#102): cascade scoring-level z-norm — replace raw cosine with calibrated z-score in scoreCandidate

tuky191:
## Summary

Closes the structural failure mode behind #102 — wrong-person reacquires admitted on raw cosine alone, plus the cross-category leak (person-locked admitting dog/remote) that surfaced when running the full on-device suite.

Three coordinated changes:

1. **Scoring-level z-norm in person paths** (`hasFace`, `hasReId`) — replace raw cosine in the cascade weighted-sum with `znormToScore(z-score)` mapped to [0,1] via sigmoid. Cold-start falls back to raw cosine when cohort < 5.
2. **`hasAppearance` (non-person) branch left on raw cosine** — Phase 2 calibration was derived from person-vs-person reacquires; non-person scenes have a cohort populated from cross-category negatives so the z-distribution is uncalibrated.
3. **Hard person/non-person category gate** — drop `labelOverride`; `lockedIsPerson != candidateIsPerson` is now an unconditional reject. The category gate's purpose is to prevent person↔non-person identity confusion; letting any embedding signal bypass it defeats that purpose.

## What this fixes

**Wrong-person reacquires on raw cosine** — before: cascade summed raw cosines where reId carried ~60% weight. An impostor at typical OSNet cosine=0.65 contributed `0.65 × 0.6 = 0.39`, almost passing the 0.45 admit threshold on a single weak signal. Now: same impostor at z=0 maps to 0.27 → contribution 0.16. Same-person at z=2 maps to 0.73 → contribution 0.44.

**Cross-category leaks** — before: `labelOverride` (z ≥ 1.5 OR raw cosine ≥ 0.7) was admitting person candidates for non-person locks (and vice versa) because a totally-different category visual is "far from impostor mean" by construction (high z, but for the wrong reason). Now: hard reject regardless of embedding evidence.

## Calibration

Sigmoid mapping `znormToScore(z) = 1 / (1 + exp(-(z - 1) × 1))` per Phase 2 calibration (`project_102_reid_score_normalization.md`):

| z-score | output | calibration class |
|---|---|---|
| -1 | 0.12 | impostor (weak contribution) |
| 0 | 0.27 | impostor at distribution mean |
| 1 | 0.50 | boundary |
| 1.5 | 0.62 | minimum same-person observed |
| 2 | 0.73 | typical same-person |
| 3 | 0.88 | strong same-person |

Sigmoid centered at z=1 because Phase 2 measurements showed clean separation: same-person z ≥ +1.5, wrong-person z ≤ +0.85.

## Bundled bug fix

The `hasReId` branch was double-counting OSNet (the `appearanceScore` alias equalled `reIdScore` when `hasReId` so OSNet got `effReIdW + baseEmbW` weight). Routed secondary term through `mnv3Calibrated` so MNV3 actually fills that slot.

## Verified — full on-device suite (21 tests)

| Test | Result | Notes |
|---|---|---|
| `walking_outdoor_diagnostic` | log-only — improved | Pre-PR: 39% / 6 reacqs / **timeout**. Now: 4-6 reacqs / no timeout (variance) |
| `boy_indoor_wife_swap_reacquires_correctly` | **pass** | Was failing on cross-category leak (`remote@F1211`, `dog@F77`). Hard gate fixed. |
| `boy_indoor_wife_swap_tracking_rate` | pass | 80% (baseline 83-86%) |
| `chair_living_room_tracking_rate` | fail (63%, baseline 76-77%) | **Improved 25% → 63%** — partial #110 fix from removing the labelOverride bypass. Remaining 13-point gap is the original #110 regression, not Phase 4. |
| `mouse_desk_rotation_tracking_rate` | fail (19%, baseline 29-34%) | Within known noise band; non-Phase-4. |
| All 17 other tests | pass | |

JVM unit tests: 295 pass. The one previously asserting the old override behavior (`strong embedding overrides person-not-person gate`) was rewritten to assert the new hard-gate behavior.

## Calibration on the walking session (multi-person, cohort matures)

All 5 same-person reacquires at `znOsnet 1.8-2.8` → mapped to 0.73-0.86 → admitted. One candidate at `znOsnet=-0.469` correctly rejected (would have been admitted pre-fix at raw `reId=0.445 × 0.6 = 0.27` plus other signals).

## What's NOT in this PR

- **Cold-start coverage** — single-person scenes still rely on raw cosine until the impostor cohort matures (MIN_COHORT_FOR_ZNORM=5). Phase 2 found that frozen-pool backfill broke z-score distributions, so cold-start improvement requires a separate calibration design.
- **#116 orientation half** — pipeline-wide rotation fix, separate PR.
- **#110 chair regression residual** — 63% → 76-77% gap remains; bisection target separate from this work. Comment posted on #110 noting the partial improvement.

## Test plan

- [x] `./gradlew testDebugUnitTest` — 295 unit tests pass
- [x] `./gradlew assembleDebug assembleDebugAndroidTest` — clean build
- [x] Full on-device suite — 18/21 pass; 2 pre-existing failures (chair #110 partial, mouse variance); cross-category leak fixed.
- [x] Inspected session.log — z-norm-driven reacquire decisions visible (znOsnet 1.8-2.8 admitted, -0.5 rejected)

### DIFF
--- a/app/src/main/java/com/haptictrack/tracking/ReacquisitionEngine.kt
+++ b/app/src/main/java/com/haptictrack/tracking/ReacquisitionEngine.kt
@@ -2,6 +2,7 @@ package com.haptictrack.tracking
 
 import android.graphics.RectF
 import android.util.Log
+import kotlin.math.exp
 
 /**
  * Pure logic for re-acquiring a lost tracking target.
@@ -68,6 +69,18 @@ class ReacquisitionEngine(
          *  Phase 2). Cap σ at this floor so the override decision can't be
          *  driven by a degenerate cohort. */
         const val Z_SIGMA_FLOOR = 0.05f
+
+        // --- Phase 4 (#102): scoring-level z-norm constants ---
+        /** Sigmoid midpoint when mapping a z-score into [0,1] for the cascade
+         *  scoring formula. Phase 2 calibration on real reacquires showed:
+         *  same-person z ≥ +1.5, wrong-person z ≤ +0.85. The clean separator
+         *  is z ≈ +1.0, so map z=1.0 → output 0.5 (the sigmoid midpoint). */
+        const val Z_SCORE_MIDPOINT = 1.0f
+        /** Sigmoid steepness. With scale=1, z=0 → 0.27, z=1 → 0.5, z=2 → 0.73,
+         *  z=3 → 0.88. That keeps the calibration band (-1..+3) inside the
+         *  steep middle of the sigmoid where the score actually moves with
+         *  the z-score. Higher scale = sharper boundary; lower = more lenient. */
+        const val Z_SCORE_SCALE = 1.0f
         /** Minimum embedding similarity to consider a candidate at all.
          *  If the primary embedder says the candidate is a different object (sim < this),
          *  no amount of re-ID, attributes, or color can rescue it. */
@@ -353,6 +366,31 @@ class ReacquisitionEngine(
     val mnv3LiveStats: ImpostorStats? get() { recomputeLiveStatsIfNeeded(); return _mnv3LiveStats }
     val osnetLiveStats: ImpostorStats? get() { recomputeLiveStatsIfNeeded(); return _osnetLiveStats }
     val faceLiveStats: ImpostorStats? get() { recomputeLiveStatsIfNeeded(); return _faceLiveStats }
+
+    /**
+     * Phase 4 (#102): map a z-score into [0,1] for the cascade scoring formula
+     * via a sigmoid centered at [Z_SCORE_MIDPOINT]. The cascade ranks candidates
+     * with a weighted sum of [0,1]-valued signals; raw cosine is in that range
+     * naturally but isn't calibrated against the impostor distribution. A
+     * z-score IS calibrated but unbounded — the sigmoid maps it into [0,1] so
+     * "above the impostor distribution" pushes the contribution up while
+     * "inside the impostor distribution" keeps it low.
+     *
+     * With midpoint=1, scale=1: z=-1 → 0.12, z=0 → 0.27, z=1 → 0.5, z=2 → 0.73,
+     * z=3 → 0.88. Calibration band (-1..+3) sits inside the steep region.
+     */
+    private fun znormToScore(z: Float): Float =
+        1f / (1f + exp(-(z - Z_SCORE_MIDPOINT) * Z_SCORE_SCALE))
+
+    /**
+     * Phase 4 (#102): scoring-level z-norm. Returns the calibrated score for
+     * [rawCosine] — z-norm-mapped via [znormToScore] when [zScore] is available
+     * (cohort ≥ [MIN_COHORT_FOR_ZNORM]), otherwise the raw cosine itself as a
+     * cold-start fallback. Both paths return a [0,1] value so the cascade's
+     * weighted sum is unaffected.
+     */
+    private fun calibratedFromZ(rawCosine: Float, zScore: Float?): Float =
+        if (zScore != null) znormToScore(zScore) else rawCosine
 
     /**
      * Pick the appropriate z-score for [candidate] mirroring [effectiveAppearanceSim]'s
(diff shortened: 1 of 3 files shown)


tuky191 (commented):
## Code Review — PR #119 (Phase 4 cascade scoring-level z-norm, #102)

**Scope:** Three coordinated changes: (1) replace raw cosine with sigmoid-mapped z-score in the `hasFace` and `hasReId` scoring branches, (2) hard person/non-person category gate (remove `labelOverride`), (3) fix OSNet double-counting bug in `hasReId` branch.

295 unit tests pass. 18/21 on-device tests pass (2 pre-existing, 1 partial #110 improvement).

---

### Medium

**1. No dedicated unit test for the z-norm scoring path itself**
The existing `ZNormTest` covers the z-score computation and the override gates, but there's no test that verifies the scoring formula produces different final scores when z-norm is active vs inactive. For example: lock on person, add ≥5 scene negatives, score a candidate — verify the returned score uses `znormToScore` rather than raw cosine. The current test change only verifies the hard category gate (label override removed). The z-norm-in-scoring is the headline change and it's only validated through on-device replay, not JVM.

Consider adding a test like: "candidate at raw reIdScore=0.65 scores lower with z-norm active (z≈0, sigmoid≈0.27) than with raw cosine (0.65)" — that's the structural fix this PR delivers.

---

### Low

**2. `Z_LABEL_OVERRIDE_THRESHOLD` is declared but no longer used**
Line 58: `const val Z_LABEL_OVERRIDE_THRESHOLD = 1.5f` — still defined in the companion object. The `labelOverride` code that consumed it is removed. The constant is only referenced in two comments (line 44 doc, line 1051 "was removed" comment). Dead code — should be removed, and the doc reference on `APPEARANCE_OVERRIDE_THRESHOLD` (line 44) updated since it points to the now-unused constant.

**3. `APPEARANCE_OVERRIDE_THRESHOLD` doc says "See Z_LABEL_OVERRIDE_THRESHOLD for the calibrated path"**
Line 44: this doc is now misleading since the label override no longer exists. The calibrated path for `APPEARANCE_OVERRIDE_THRESHOLD` is now only the geometric override and tentative bypass — the label override is gone entirely.

---

### Observations

- The sigmoid mapping is clean: `znormToScore(z) = 1 / (1 + exp(-(z - 1) * 1))`. Midpoint at z=1, the Phase 2 boundary. z=-1 → 0.12, z=0 → 0.27, z=1 → 0.5, z=2 → 0.73, z=3 → 0.88. The calibration band sits in the steep region where the output is responsive to z changes.

- The `calibratedFromZ` fallback (`rawCosine` when z is null) is correct — cold-start behavior is preserved until the cohort matures.

- The OSNet double-counting fix is real and well-explained: `appearanceScore` equalled `reIdScore` when `hasReId`, so the `hasReId` branch was `reIdScore * effReIdW + reIdScore * baseEmbW` — OSNet counted at 60% weight instead of the intended 40%. Now it's `reIdCalibrated * effReIdW + mnv3Calibrated * baseEmbW`.

- The hard category gate is the right call. The comment (lines 114-123) gives a clear explanation of why the override was wrong: a cross-category visual is "far from impostor mean" by construction (high z, but meaningless for identity). Label flicker within a category (bowl↔potted plant) is handled by the embedding gate above.

- Weight math checks out for both `hasFace` and `hasReId` tiers — base weights sum to ≤1.0 with the unused remainder absorbed by the primary signal.

- The `hasAppearance` branch intentionally staying on raw cosine is well-reasoned — Phase 2 calibration was person-only, and the non-person negative distribution is structurally different.

---

### Verdict

The structural fix this PR delivers (impostor at z=0 → contribution 0.16 vs previous raw 0.39) is the right answer to #102's core problem. The hard category gate and the double-counting fix are both clean. The one medium (missing unit test for z-norm scoring) is worth adding before merge; the lows are cleanup. Good to merge with that addition.

tuky191 (commented):
## Follow-up review — PR #119, commits 65235339 + f9f36b67 + 171a1b4e

All three findings addressed:

1. **Medium (no z-norm scoring unit test)** ✅ — New test `hasReId scoring uses calibrated z-score, not raw cosine, when cohort matures`. Runs same candidate through `scoreCandidate` with 4 impostors (cold-start → raw cosine) vs 5 (z-norm active). Same box, same raw OSNet cosine 0.65 — isolates the z-norm delta. Asserts calibrated score is ≥0.10 lower. Clean proof of the structural fix.

2. **Low (Z_LABEL_OVERRIDE_THRESHOLD dead code)** ✅ — Constant removed. Associated doc references updated.

3. **Low (stale APPEARANCE_OVERRIDE_THRESHOLD doc)** ✅ — Rewritten: now correctly states it only feeds geometric and tentative-confirmation override paths, and that the category gate is hard.

Good to merge.

Issue #131, opened 2026-04-30
Build offline model benchmark infrastructure (prerequisite for Phases 3, 5)


tuky191:
## Context

Phases 3 (#126) and 5 (#128) of the model stack modernization epic (#129) both require offline benchmark infrastructure before writing migration code. This is a discrete piece of work that gates two phases — it needs its own issue so it doesn't fall into the "everyone's responsibility = nobody's responsibility" gap.

## What to build

A tool that runs a candidate model against the captured scenario corpus and produces a clear "better/worse/same" verdict that isn't threshold-tunable.

### Inputs
- A TFLite/ONNX model (the candidate)
- The captured scenario corpus: real device captures from `/sdcard/Android/data/com.haptictrack/files/debug_frames/session_*/` (scenario.json + raw PNGs)
- The incumbent model (for comparison)

### Process
1. For each scenario, extract all crops that were used for embedding (lock crops, candidate crops, negative crops)
2. Run both candidate and incumbent models on the same crops
3. Compute cosine similarity matrices for each scenario

### Outputs
- **Same-vs-different cosine distributions** per scenario, per model: histograms showing the separation between same-object and different-object pairs
- **Gap metric**: quantified separation (e.g., mean_same - mean_different, or d-prime) per model per scenario
- **Pass/fail verdict**: does the candidate produce a wider or narrower same/different gap than the incumbent on our scenes?
- Summary table across all scenarios

## Design considerations

- Should run offline (laptop/desktop), not on device — we want to evaluate models before going through TFLite conversion + device deployment
- Python-based, lives in `tools/` alongside existing `benchmark_embeddings.py` and `test_model_quality.py`
- Should support both TFLite models (via `tflite_runtime` or `mediapipe`) and PyTorch/ONNX models (for candidates that haven't been converted yet)
- Reuse existing `tools/` infrastructure where possible

## Relationship to existing tools

- `tools/benchmark_embeddings.py` — already runs models on crops and prints similarity matrices. This issue extends it to comparative evaluation (candidate vs incumbent) and corpus-level aggregation.
- `tools/test_model_quality.py` — pytest-based regression checks. The offline benchmark produces data; the test suite can consume it.

## Used by

- #126 Phase 3 (MobileCLIP2-S0 vs MobileNetV3 — exit criterion: wider same/different gap on our scenes)
- #128 Phase 5 (LiteTrack candidate — measure VT loss rate and box-growth on chair, mouse, bottle scenarios)
[...]

Issue #128, opened 2026-04-30
Phase 5: Replace VitTracker with LiteTrack-B4 or HiT


tuky191:
## Context

VitTracker (0.7MB, OpenCV DNN, CPU) is the weakest model in accuracy terms. Multiple edge-optimized trackers now offer better tracking with acceptable size/latency.

**Important caveat**: the +33% tracking accuracy claim is from Jetson benchmarks on LaSOT/GOT-10k, not Adreno 740 on handheld footage. The actual VT pain points are box growth and small-object failure — a better AUC may or may not fix those. Gate on offline measurement against our scenarios.

## Current state

- **VitTracker** (0.7MB ONNX, ~5ms CPU, 48.6 AUC on LaSOT)
- Integrated via OpenCV `TrackerVit` wrapper
- Provides confidence score (critical for drift detection)
- Known issues: regression head lets boxes grow (mitigated by #120 reseat), dies fast on small/transparent objects (#59)

## Candidates

| Tracker | Size | Edge FPS | Accuracy | Export | Confidence? |
|---|---|---|---|---|---|
| VitTracker (current) | 0.7MB | ~200 CPU | 48.6 AUC | ONNX/OpenCV | Yes |
| **LiteTrack-B4** | ~5MB | 44 Jetson Orin NX | 65.2 AO GOT-10k | ONNX | Needs verification |
| **HiT / DyHiT** | ~5-8MB | 61-111 Jetson AGX | 62-65% AUC LaSOT | Not documented | Needs verification |
| **MixFormerV2-S** | ~4MB | 45 CPU | 60.6% AUC | ONNX | Yes (distillation) |
| **UETrack-T** | ~6MB | 77 Jetson AGX | competitive | Code only (Mar 2026) | Unknown |
| **MVT** | ~5MB | 70 CPU | competitive | ONNX + TRT | Uses MobileViT backbone |

## Before committing to a swap

1. **Check OpenCV DNN compatibility**: can the candidate ONNX models be loaded via OpenCV DNN (which we already have)? Strongly prefer OpenCV-DNN-compatible options — avoids adding ONNX Runtime as a new dependency.
2. **Run candidate against our scenarios offline**: chair_living_room, mouse_desk_rotation, bottle captures. Measure VT loss rate, box-growth events, and confidence-score quality. The real metric is "fewer scenarios timing out due to VT failure," not LaSOT AUC.
[...]

Issue #126, opened 2026-04-30
Phase 3: Consolidate embedders — MobileNetV3 + MobileViTv2 ×2 → MobileCLIP2-S0


tuky191:
## Context

Three models currently handle generic embedding + classification:
1. **MobileNetV3 Large** (10.4MB, 1280-dim) — primary gallery embedder
2. **MobileViTv2-0.75 embedder** (9.8MB) — alternative embedder for canonical-crop work
3. **MobileViTv2-0.75 classifier** (11.3MB, 1000 logits) — online classifier for tentative confirmation

Total: **31.5MB** for closely related functionality.

## This issue should be split into 3 sub-issues

These are three independent decisions with different risk profiles:

### Sub-issue 3a: Audit MobileViTv2 embedder usage
- Confirm whether the MobileViTv2-0.75 embedder is active in any production code path, or only used by audit/benchmark tooling (#91, #100)
- If near-dead → drop it (free 9.8MB, trivially safe, no calibration impact)
- **Effort**: 1 hour code audit

### Sub-issue 3b: Replace CropClassifier (MobileViTv2 classifier)
- Replace `CropClassifier` (MobileViTv2 classifier, 11.3MB) with zero-shot classification via CLIP text embeddings, or alternative
- Lower risk than replacing the primary embedder — `CropClassifier` is used only for tentative confirmation, not the cascade gates
- **Effort**: 1 week

### Sub-issue 3c: Replace MobileNetV3 (primary embedder) with MobileCLIP2-S0
- **This is the calibration-heavy, highest-risk change in the epic**
- Invalidates all calibration work from #102/#112/#113/#119/#120: `lockSelfFloor`, z-norm cohorts, `APPEARANCE_OVERRIDE_THRESHOLD`, gallery accumulator gate, roster floors
- CLIP space behaves differently from ImageNet-pretrained space — CLIP embeddings tend to be tightly clustered for "this is a photograph of an object" regardless of which object, compressing the same/different gap

## Exit criterion (must be answered BEFORE writing migration code)

The validation criterion "all replay tests pass with recalibrated thresholds" is circular — you can tune until they pass. Instead, the following must be demonstrated **offline on the captured scenario corpus** (real device captures, not unit-test fixtures):

1. **Is MobileViTv2 embedder actually used?** (Quick code audit → Sub-issue 3a)
[...]

PR #148, opened 2026-05-01
Offline model benchmark infrastructure (#131)

tuky191:
## Summary
- Comparative benchmark tool that runs incumbent vs candidate embedding models on the scenario corpus
- Produces a non-threshold-tunable "better/worse/same" verdict based on same_p10 - diff_p90 gap and d-prime
- Supports three model backends: `mediapipe:`, `tflite:`, `onnx:`
- Pytest gate tests that block model swaps if the candidate regresses

## Usage
```bash
cd tools

# CLI comparison
.venv/bin/python benchmark_compare.py \
    --incumbent mediapipe:../app/src/main/assets/mobilenet_v3_large_embedder.tflite \
    --candidate tflite:path/to/candidate.tflite \
    --output results.json

# Pytest gate (from pre-computed JSON)
BENCHMARK_RESULTS=results.json .venv/bin/pytest test_benchmark_compare.py -v

# Pytest gate (inline, re-runs inference)
.venv/bin/pytest test_benchmark_compare.py -v \
    --incumbent mediapipe:../app/src/main/assets/mobilenet_v3_large_embedder.tflite \
    --candidate tflite:path/to/candidate.tflite
```

## Verified
- MobileNetV3 vs MobileNetV3 → SAME (gap delta = +0.000)
- MobileNetV3 vs OSNet → BETTER (gap delta = +0.234, d-prime -1.12 → +3.70)
- OSNet vs MobileNetV3 → WORSE — all 4 gate tests correctly fail

## Test plan
- [x] Same-model comparison produces SAME verdict
- [x] Better candidate passes all 4 gate tests
- [x] Worse candidate fails all 4 gate tests with clear error messages
- [x] Existing test_model_quality.py unaffected (same 2 pre-existing failures on two_cars)
- [x] Both env var and inline pytest invocation modes work

Prerequisite for Phase 3c (#126) and Phase 5 (#128).

### DIFF
--- a/tools/benchmark_compare.py
+++ b/tools/benchmark_compare.py
@@ -129,7 +129,7 @@ def _resize_letterbox(img: Image.Image, target_w: int, target_h: int) -> Image.I
     new_w = max(1, int(src_w * scale))
     new_h = max(1, int(src_h * scale))
     resized = img.resize((new_w, new_h), Image.BILINEAR)
-    canvas = Image.new("RGB", (target_w, target_h), (114, 114, 114))
+    canvas = Image.new("RGB", (target_w, target_h), (128, 128, 128))
     canvas.paste(resized, ((target_w - new_w) // 2, (target_h - new_h) // 2))
     return canvas
 
@@ -160,8 +160,14 @@ def build_tflite_embedder(model_path: str) -> EmbedFn:
         import tflite_runtime.interpreter as tflite
         Interpreter = tflite.Interpreter
     except ImportError:
-        import tensorflow as tf
-        Interpreter = tf.lite.Interpreter
+        try:
+            import tensorflow as tf
+            Interpreter = tf.lite.Interpreter
+        except ImportError:
+            raise ImportError(
+                "tflite backend requires tflite-runtime or tensorflow: "
+                "pip install tflite-runtime  (or)  pip install tensorflow"
+            )
 
     interpreter = Interpreter(model_path=model_path)
     interpreter.allocate_tensors()
@@ -191,11 +197,17 @@ def build_tflite_embedder(model_path: str) -> EmbedFn:
             feat = feat / n
         return feat
 
+    embed._close = lambda: None
     return embed
 
 
 def build_onnx_embedder(model_path: str) -> EmbedFn:
-    import onnxruntime as ort
+    try:
+        import onnxruntime as ort
+    except ImportError:
+        raise ImportError(
+            "onnx backend requires onnxruntime: pip install onnxruntime"
+        )
 
     sess = ort.InferenceSession(model_path, providers=["CPUExecutionProvider"])
     inp = sess.get_inputs()[0]
@@ -222,6 +234,7 @@ def build_onnx_embedder(model_path: str) -> EmbedFn:
             feat = feat / n
         return feat
 
+    embed._close = lambda: None
     return embed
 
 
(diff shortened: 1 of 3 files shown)


tuky191 (commented):
## PR #148 Review: Offline model benchmark infrastructure (#131)

**Files:** 3 new (`benchmark_compare.py`, `conftest.py`, `test_benchmark_compare.py`) — 674 additions

---

### What it does

Adds a comparative benchmark CLI that runs incumbent vs candidate embedding models on the annotated scenario corpus (`test_fixtures/`), measures same-vs-different cosine gap (`same_p10 − diff_p90`) and d-prime, and produces a structural "BETTER / WORSE / SAME" verdict. Three model backends (mediapipe, tflite, onnx) for evaluating candidates before integration. Pytest gate tests consume the JSON output and block model swaps that regress.

Good design choices:
- Gap metric uses percentiles (p10/p90), not means — robust to outliers
- d-prime is the right signal-detection metric for this task
- Session-scoped pytest fixture avoids re-running inference per test
- Two invocation modes (pre-computed JSON or inline) cover CI and interactive use
- Letterbox resize in tflite/onnx backends preserves aspect ratio correctly

---

### Findings

**[Low] Optional backend imports give unhelpful errors**
`build_tflite_embedder` (line 165) and `build_onnx_embedder` (line 204) import `tflite_runtime`/`tensorflow` and `onnxruntime` respectively — none of which are in `requirements.txt`. A user passing `tflite:path` will get a raw `ImportError` with no guidance. Consider wrapping with a friendlier message:
```python
try:
    import onnxruntime as ort
except ImportError:
    raise ImportError("onnx backend requires: pip install onnxruntime")
```
Or document the optional deps in the module docstring.

**[Low] `_crop_box` duplicates `benchmark_embeddings.py` crop logic, but better**
`benchmark_compare.py:_crop_box` (line 119) adds proper edge clamping that the original `crop_and_embed()` in `benchmark_embeddings.py:66-76` lacks. Not a merge blocker — but worth knowing the older code has a latent edge-crop bug, and if you later consolidate, the new version should win.

**[Nit] Letterbox pad color (114,114,114) is YOLOv5 convention**
Line 138 uses gray fill `(114, 114, 114)` which is the YOLO standard. For ImageNet-normalized models the padding becomes a non-zero signal (`≈ -0.73` after normalization). Impact is minimal since padding area is small for typical crops, but `(128, 128, 128)` or even `(0, 0, 0)` would be more neutral. No action needed unless you see artifacts on very small crops.

**[Nit] `conftest.py` adds global pytest options visible to all test files**
The `--incumbent` and `--candidate` options are registered globally via `conftest.py`, so `test_model_quality.py` (which uses `MODEL_PATH` env var) also sees them. Harmless since those options default to `None` and the other test file doesn't read them — just noting the two model-specification patterns now coexist (`env var MODEL_PATH` vs `--incumbent`/`--candidate` CLI args).

**[Nit] Resource cleanup inconsistency**
`build_mediapipe_embedder` attaches `_close` (line 161) and `run_model` calls it (line 391). The tflite and onnx builders don't attach `_close`, so their interpreters/sessions live until GC. Fine for a CLI tool, just inconsistent.

---

### Verdict

**Good to merge.** Clean infrastructure — proper metrics, good separation of concerns, two invocation modes for flexibility. The optional-backend import UX is the only thing worth addressing before the tflite/onnx paths see real use (and Phase 3c will exercise `tflite:` immediately). The rest are nits.

PR #150, opened 2026-05-01
Enable ISP stabilization + couple tracking with recording

tuky191:
## Summary

- **CameraX preview stabilization**: enables `setPreviewStabilizationEnabled(true)` when the device supports it. Stabilization happens at the ISP/HAL level — both analysis frames (SurfaceTexture → tracking pipeline) and video recording get stabilized pixels with zero added latency. Confirmed working on Poco F4.
- **Track+record coupling**: stopping recording now auto-clears tracking. Combined with lock auto-starting recording (PR #111), tracking and recording are always paired. Volume-down cycle unchanged.

## Test plan

- [x] Build and deploy
- [x] Logcat confirms `"Preview stabilization enabled (ISP-level EIS)"` on Poco F4
- [ ] Device test: handheld tracking with stabilization — compare zoom stability vs. before
- [ ] Device test: stop recording clears lock
- [ ] Device test: volume-down cycle still works (lock+record → stop+clear)

### DIFF
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -40,6 +40,9 @@ class CameraManager(private val context: Context) {
     /** Detected optical zoom limit from physical camera focal lengths. */
     private var opticalZoomMax: Float = DEFAULT_OPTICAL_MAX
 
+    /** Software gyro-based EIS, stacks on top of ISP stabilization. */
+    val gyroStabilizer = GyroStabilizer(context)
+
     /** Current lens facing — back by default. */
     var isFrontCamera: Boolean = false
         private set
@@ -84,11 +87,13 @@ class CameraManager(private val context: Context) {
 
     init {
         opticalZoomMax = detectOpticalZoomMax()
+        gyroStabilizer.readCameraIntrinsics(context)
     }
 
     fun startCamera(lifecycleOwner: LifecycleOwner, previewView: PreviewView) {
         lifecycleOwnerRef = lifecycleOwner
         previewViewRef = previewView
+        gyroStabilizer.start()
         val providerFuture = ProcessCameraProvider.getInstance(context)
         providerFuture.addListener({
             cameraProvider = providerFuture.get()
@@ -144,7 +149,7 @@ class CameraManager(private val context: Context) {
         preview = previewBuilder.build()
 
         // Always route Preview to SurfaceTextureFrameReader for fast off-thread frame capture.
-        // 2-stream binding (preview + video) enables 4K recording.
+        // 2-stream binding (preview + video) — FHD when stabilized, 4K otherwise.
         preview.surfaceProvider = Preview.SurfaceProvider { request ->
             val inputSize = request.resolution // camera's native buffer size (landscape, e.g. 1600x1200)
             // Output in portrait at analysis resolution.
@@ -164,7 +169,8 @@ class CameraManager(private val context: Context) {
                 outputWidth = outW,
                 outputHeight = outH,
                 onFrame = { bitmap -> onAnalysisFrame?.invoke(bitmap) },
-                onViewfinderFrame = { bitmap -> onViewfinderFrame?.invoke(bitmap) }
+                onViewfinderFrame = { bitmap -> onViewfinderFrame?.invoke(bitmap) },
+                stabMatrixProvider = { gyroStabilizer.getMatrix() }
             )
             val readerSurface = reader.start()
             frameReader = reader
@@ -242,6 +248,7 @@ class CameraManager(private val context: Context) {
     }
 
     fun shutdown() {
+        gyroStabilizer.stop()
         frameReader?.stop()
         frameReader = null
         cameraProvider?.unbindAll()
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -74,7 +74,7 @@ class CameraManager(private val context: Context) {
     /** Callback for viewfinder display frames from SurfaceTexture (GL thread, ~29fps). */
     var onViewfinderFrame: ((android.graphics.Bitmap) -> Unit)? = null
 
-    var preview = Preview.Builder().build()
+    lateinit var preview: Preview
         private set
 
     private val videoExecutor = Executors.newSingleThreadExecutor()
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -2,8 +2,10 @@ package com.haptictrack.camera
 
 import android.content.Context
 import android.hardware.camera2.CameraCharacteristics
+import android.hardware.camera2.CaptureRequest
 import android.hardware.camera2.CameraManager as Camera2Manager
 import android.util.Log
+import androidx.camera.camera2.interop.Camera2Interop
 import androidx.camera.core.CameraControl
 import androidx.camera.core.CameraInfo
 import androidx.camera.core.CameraSelector
@@ -150,9 +152,8 @@ class CameraManager(private val context: Context) {
         val selector = if (isFrontCamera) CameraSelector.DEFAULT_FRONT_CAMERA
                        else CameraSelector.DEFAULT_BACK_CAMERA
 
-        // ISP stabilization and gyro EIS are independently togglable.
         val previewBuilder = Preview.Builder()
-        if (ispStabilizationEnabled) {
+        if (ispStabilizationEnabled && !gyroStabilizer.enabled) {
             try {
                 val caps = Preview.getPreviewCapabilities(provider.getCameraInfo(selector))
                 if (caps.isStabilizationSupported) {
@@ -164,8 +165,19 @@ class CameraManager(private val context: Context) {
             } catch (e: Exception) {
                 Log.w(TAG, "Failed to query stabilization caps: ${e.message}")
             }
+        } else if (ispStabilizationEnabled && gyroStabilizer.enabled) {
+            Log.i(TAG, "ISP stabilization OFF (gyro EIS takes over)")
         } else {
             Log.i(TAG, "ISP stabilization OFF (user toggle)")
+        }
+        @Suppress("UnsafeOptInUsageError")
+        if (gyroStabilizer.enabled) {
+            Camera2Interop.Extender(previewBuilder)
+                .setCaptureRequestOption(
+                    CaptureRequest.LENS_OPTICAL_STABILIZATION_MODE,
+                    CaptureRequest.LENS_OPTICAL_STABILIZATION_MODE_OFF
+                )
+            Log.i(TAG, "OIS disabled (gyro EIS handles stabilization)")
         }
         gyroStabilizer.readCameraIntrinsics(context, frontFacing = isFrontCamera)
         Log.i(TAG, "Gyro EIS ${if (gyroStabilizer.enabled) "ON" else "OFF"}")
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -62,7 +62,8 @@ class CameraManager(private val context: Context) {
     /** Callback for viewfinder display frames from SurfaceTexture (GL thread, ~29fps). */
     var onViewfinderFrame: ((android.graphics.Bitmap) -> Unit)? = null
 
-    val preview = Preview.Builder().build()
+    var preview = Preview.Builder().build()
+        private set
 
     private val videoExecutor = Executors.newSingleThreadExecutor()
     var videoCapture = createVideoCapture()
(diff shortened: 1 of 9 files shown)


tuky191 (commented):
## PR #150 Review: ISP stabilization + gyro EIS + track-record coupling

**Files:** 5 changed (421 additions, 7 deletions)  
**New class:** `GyroStabilizer` (270 lines), `GyroStabilizerTest`  
**Modified:** `CameraManager`, `SurfaceTextureFrameReader`, `CameraViewModel`

Three changes bundled:
1. CameraX `setPreviewStabilizationEnabled(true)` when hardware supports it
2. Gyro-based software EIS (Gyroflow-style): quaternion SLERP smoothing → homography correction → GL shader warp
3. Stop-recording auto-clears tracking (couples with PR #111's lock-auto-records)

---

### Findings

**[High] ISP and gyro EIS will interfere when both active — the software correction assumes raw (unstabilized) frames**

The gyro stabilizer computes `correction = rawQuat * smoothedQuat.conjugate()` — the rotation from the smoothed orientation to the raw device orientation. This homography is applied to texture coordinates in the GL shader. The assumption is: the frame content sits at `rawQuat`'s orientation, so the warp compensates for the shake delta.

When ISP stabilization is enabled, this assumption breaks. The ISP has already stabilized the frame — the pixel content sits at the ISP's smoothed orientation, NOT at `rawQuat`. The software correction still computes `raw → software-smoothed`, but applies it to ISP-smoothed pixels. The result:

```
effective orientation = ISP_smooth + (raw - software_smooth)
```

If both smoothers track similar signals (they do — same gyro data, different algorithms), this roughly equals `raw` — i.e., the software EIS **undoes** the ISP's stabilization and replaces it with its own. At best, the ISP's work is wasted. At worst, the two corrections fight on frames where their smoothing windows diverge (fast pans, direction changes), producing a swimming/jello artifact.

Recommendation: either (a) disable software EIS when ISP stabilization is available (ISP alone is typically higher quality — it has rolling shutter correction, sub-pixel accuracy, and sensor-level knowledge), or (b) disable ISP stabilization and let the software EIS be the sole stabilizer. Option (a) is simpler and lower risk. The software EIS is valuable as a fallback on devices where `isStabilizationSupported == false`.

**[Medium] Software EIS warp applies to tracking pipeline analysis frames**

Both the viewfinder AND analysis frames go through `SurfaceTextureFrameReader`'s GL shader, which now includes the stabilization warp (line 395-396 in the vertex shader). This means:
- VitTracker, the detector, and embeddings all see warped pixels
- The warp changes between frames (the correction quaternion evolves with time)

For the common case (small corrections), this should actually *help* VitTracker — smoother inter-frame motion means better pixel correlation. But edge cases to watch for in device testing:
- **Frame-edge artifacts:** When the correction warp pulls texture coordinates outside [0,1], the sampler clamps to edge (pixels get stretched). Objects near the frame border could get distorted, confusing VT or the detector. The 5% `cropZoom` provides some margin but doesn't cover large corrections.
- **Template self-verification:** The template check compares embeddings of VT crops across frames. Since the warp changes between frames, the crop content shifts slightly even for a perfectly tracked object. The adaptive `lockSelfFloor` (PR #120) should absorb this, but monitor for false drift detections during large shakes.

**[Low] Camera intrinsics not updated on `switchCamera()`**

`readCameraIntrinsics` is called once in `init` (line 89) and reads the back camera's focal length. If `switchCamera()` is called, the front camera has different intrinsics but the stabilizer keeps the back camera's values. Front-camera tracking is rare in this app's use case, so low impact. Fix: call `readCameraIntrinsics` in `bindUseCases()` or after camera switch.

**[Low] Gyro EIS `slerp` is private — test duplicates the implementation**

`GyroStabilizerTest` (lines 528-542) contains its own copy of the SLERP algorithm to test it. If the production `slerp` is modified, the test copy won't track the change. The quaternion algebra tests (identity, multiplication, rotation matrix) DO test the production `Quat` class directly and would catch most math bugs. Consider either making `slerp` internal/package-private or testing it indirectly via the stabilization matrix output.

---

### Track+record coupling (commit 1)

This part is clean and correct:

- `toggleRecording()` calls `clearTracking()` after `stopRecording()` when tracking is active (line 295-297)
- Volume-down cycle is unchanged — was already stop+clear, now just routes through `toggleRecording()`
- The `onVolumeDown()` method (line 225-228) previously had `toggleRecording()` then `clearTracking()` as separate calls; now `toggleRecording()` handles both, eliminating the duplicate `clearTracking()` call

### Homography math (verified)

- K × R × K⁻¹ derivation: correct ✓
- Crop zoom: scales UV around center, `tx = 0.5*(1-z)` keeps center at (0.5, 0.5) ✓
- Column-major transpose for GL: correct ✓
- Perspective divide in shader (`stabUV.xy / stabUV.z`): correct for projective warp ✓

### Tests

8 new quaternion/rotation tests pass. Full suite (301 tests) passes.

---

### Verdict

**Do not merge as-is.** The ISP + gyro EIS stacking will interfere — see the High finding. Recommend splitting the approach: ISP stabilization ON → software EIS OFF (and vice versa), so only one correction layer is active at a time. The track+record coupling (commit 1) is clean and can merge independently.

tuky191:
**Amendment to the High finding (ISP + gyro EIS stacking):**

The interference concern is based on mathematical reasoning, not verified observation. The argument is that both correction layers use the same gyro signal but apply independently — if the ISP has already stabilized the frame content, the software EIS correction (computed from raw device orientation) would over-correct.

However, I can't verify from code alone:
- What PREVIEW_STABILIZATION actually does on Poco F4 (full EIS vs rolling shutter only)
- Whether the SurfaceTexture transform matrix absorbs ISP correction
- The real magnitude of ISP correction vs software EIS correction

**Downgraded to Medium.** The concern is structurally sound but unverified. A tripod + slow-pan device test would confirm or refute it immediately — if the output overshoots/lags the pan direction, the layers are fighting. If smooth, my assumption about ISP behavior is wrong and both layers coexist fine.

tuky191:
## Review

**Verdict:** ⚠️ APPROVE_WITH_NITS — two real correctness items worth fixing before merge.

### Critical issues
- **`StabilizationProcessor.cleanupOutput()` does not call `outputSurfaceOutput?.close()`.** If `release()` runs before CameraX sends `EVENT_REQUEST_CLOSE`, the SurfaceOutput leaks. CameraX contract requires explicit close.
- **`StabilizationProcessor.released: Boolean` is not `@Volatile`.** Written on UI thread, read on GL thread in `onFrameAvailable`. JVM may cache.
- **Correction clamping is single-pass.** Excursion is non-linear in angle (perspective from `[r6/fx, r7/fy, r8]` bottom row); the recomputed `h` is not guaranteed to satisfy the margin. Either iterate 2-3 passes (they converge quickly) or assert post-clamp excursion in tests.

### Style/idiom nits
- `OOB_WARN_COOLDOWN_NS` and `telLastWarnNs` declared but unreferenced — dead code.
- `enabled` setter races: mutates `currentMatrix` and calls `resetTelemetry()` from caller thread while sensor thread reads/writes the same fields (plain `Int`/`Double`, torn reads possible). Move telemetry reset to sensor-thread-only or `@Synchronized` the setter.
- `sessionWriter` close-vs-write race can throw `IOException` (not NPE — `?.` is safe). Guard with synchronized or swallow on closed writer.
- `var preview = Preview.Builder().build()` initial value is discarded on first `bindUseCases` — use `lateinit` or drop the field initializer.
- `setGyroStrength` magic numbers (`0.30 - 0.26 × s`, `1.05 + 0.15 × s`) — promote to named constants in `GyroStabilizer`.
- `var onSessionDir: ((java.io.File?) -> Unit)?` in `ObjectTracker.kt` — qualify import at top, not inline.
- Vertex shader skips perspective divide on `uStabMatrix` — mathematically an approximation when `r[6]/r[7]` are nonzero. Acceptable for small corrections; add `// approximation valid for small angles` comment in `StabilizationProcessor.kt:855` and `SurfaceTextureFrameReader.kt:425`.
- `gyroStabilizer.readCameraIntrinsics(context)` called in init AND in `bindUseCases` — redundant but harmless.

### Summary
Solid, well-structured Gyroflow-style EIS implementation with clean separation between `GyroStabilizer` (math+sensor) and `StabilizationProcessor` (GL plumbing). Telemetry is genuinely useful (alpha/corrDeg/clamp%/excur/gap). Two real correctness items: SurfaceOutput leak on release path and missing `@Volatile` on the GL kill flag; one math caveat on single-pass clamping. Architectural fit is good — pipeline stays 2-stream, no rebind for recording, ISP and gyro paths are mutually exclusive. Track+record coupling change in `clearTracking`/`toggleRecording` is small and consistent with PR #111.

🤖 Generated by automated review.

tuky191:
## Review findings — all addressed

**[High] ISP and gyro EIS interference** — Fixed in `f6646e1b` ("Address review: ISP/gyro mutual exclusion, intrinsics on switch, slerp dedup") and `edb66c79`. ISP stabilization is disabled when gyro EIS is active, and OIS is explicitly turned off via Camera2Interop. Only one correction layer is ever active.

**[Medium] EIS warp on analysis frames** — Acknowledged. Small corrections help VT (smoother inter-frame motion). The 12.5% crop zoom provides margin. Template self-verification uses the adaptive `lockSelfFloor` which absorbs warp-induced sim variance. Monitoring during device testing.

**[Low] Intrinsics not updated on switchCamera()** — Fixed in `f6646e1b`. `readCameraIntrinsics(context, frontFacing = isFrontCamera)` is now called in `bindUseCases()` on every camera switch. Also reads SENSOR_ORIENTATION for the device-to-sensor rotation (added in `cd10725a`).

**[Low] slerp test duplication** — Fixed in `f6646e1b`. `slerp()` is now a package-level `internal fun` that both production code and tests reference directly.

tuky191:
## Re-review

**Status vs previous review:** NEW_COMMITS, ADDRESSED_SOME
**Verdict:** ⚠️ APPROVE_WITH_NITS

### Resolved (commit `510e5ca`)
- ✅ `@Volatile` added to `StabilizationProcessor.released` (StabilizationProcessor.kt:646)
- ✅ `outputSurfaceOutput?.close()` now called in `cleanupOutput()` (StabilizationProcessor.kt:829)
- ✅ Dead `OOB_WARN_COOLDOWN_NS` / `telLastWarnNs` removed from `GyroStabilizer.kt`

### Still outstanding
- Correction clamping is still single-pass (`GyroStabilizer.kt:406-414`). For large excursions the shrunk quaternion's max-corner is not exactly `usableMargin` because the homography is non-linear in rotation angle. Comment or document the residual is fine, but it's still mathematically not a guarantee.
- `setGyroStrength` magic numbers `0.30/0.26/1.05/0.15` still inline (`CameraViewModel.kt:1166-1167`) — should be named constants matching the doc.
- Vertex shader two-step UV transform still missing the small-angle / `.xy` projection comment (`StabilizationProcessor.kt:854-855` and `SurfaceTextureFrameReader.kt:970-971`). `(uStabMatrix * vec3(uv,1)).xy` discards `w`; works only when the homography keeps `w≈1`.
- `var preview = Preview.Builder().build()` still has wasted initial value (`CameraManager.kt:49`) — `lateinit var` is the idiomatic fix.
- `enabled` setter still mutates `currentMatrix` and calls `resetTelemetry()` from caller thread while sensor thread updates the same fields (`GyroStabilizer.kt:253-254`).
- `sessionWriter` close-vs-write race remains (`endSessionLog()` and `println` on different threads).

### New findings
- `CameraManager.shutdown()` calls `gyroStabilizer.stop()` but `bindUseCases` calls `gyroStabilizer.start()` only inside `startCamera()` — switching cameras via `bindUseCases` does not restart the sensor listener. Benign today since `start()` happens once per process; worth a one-line comment.
- `toggleIspStabilization()` triggers `cameraManager.rebind()` from UI thread; `rebind()` then re-reads `gyroStabilizer.enabled`. If a user double-taps quickly, prior stabProcessor is released and a new one created mid-binding — CameraX likely tolerates it but no guard.

### Summary
The three Critical items from the previous review are addressed and look correct. The Important style/idiom items are mostly still present but are non-blocking. Big-picture wiring (ISP+gyro toggle, OIS mutual exclusion, video stabilization through CameraEffect, session log lifecycle via `onSessionDir`) is coherent. Recommend merging and tracking the residual nits as cleanup tasks.

🤖 Generated by automated review.

tuky191:
## 3rd-pass review

**Status vs previous re-review:** NEW_COMMITS, ADDRESSED_MOST (commit `905f91a`)
**Verdict:** ⚠️ APPROVE_WITH_NITS

### Resolved
- ✅ **Item 1** — Single-pass SLERP clamping now documented (`GyroStabilizer.kt:222-226`) — sub-pixel residual is acknowledged as an acceptable trade-off rather than iteratively fixed. Fine.
- ✅ **Item 2** — Named constants added (`GYRO_TC_MAX/RANGE`, `GYRO_CROP_MIN/RANGE` in `CameraViewModel.kt`).
- ✅ **Item 3** — `.xy` perspective-divide comment added in both shaders (`StabilizationProcessor.kt`, `SurfaceTextureFrameReader.kt`).
- ✅ **Item 4** — `lateinit var preview: Preview` replaces wasted initial build (`CameraManager.kt:77`).
- ✅ **Item 5** — `enabled` setter race fixed: `resetTelemetry()` removed; only AtomicReference write + `@Volatile initialized=false` remain. No cross-thread float accumulator race.
- ✅ **Item 6 (partial)** — write at line 258 wrapped in try/catch.

### Still outstanding
- **Item 6 (residual)** — the `WARN` write at `GyroStabilizer.kt:196` (`sessionWriter?.println("... WARN ...")`) is still unguarded — same close-vs-write race window. Trivial fix; either guard it too or extract a `safeWriteLog(line)` helper.
- **Item 7** — benign, unchanged (sensor listener not restarted on `bindUseCases`).
- **Item 8** — ISP toggle mid-binding race, unchanged.

### New findings
- `lateinit var preview` will throw `UninitializedPropertyAccessException` if any external caller reads `cameraManager.preview` before `startCamera`/`bindUseCases` runs. The previous default-built `Preview` masked this. In practice consumers only access it after binding, so likely safe — worth a brief audit.
- `initialized = false` in the `enabled=false` setter is a behavior change — disabling now forces re-initialization of `smoothedQuat` from raw on next sample. Looks intentional and correct (avoids a stale smoothed quat carrying over after disable/enable cycles).

### Summary
The new commit cleanly addresses 5 of 6 nits from the prior re-review. One leftover unguarded write at line 196 is the only residual inconsistency. Safe to approve with the optional follow-up to mirror the try/catch on the WARN log line.

🤖 Generated by automated review.

Issue #162, opened 2026-05-03
EIS: horizon lock using gravity vector


tuky191:
**Epic:** #159

## Problem
Current EIS removes translational shake but preserves roll. During walking, the phone rocks side-to-side, creating a tilting horizon that looks unnatural even when jitter is removed.

## Proposed solution
Extract the gravity-aligned "up" direction from `TYPE_GAME_ROTATION_VECTOR` (already fused with accelerometer). Compute residual roll angle between the smoothed orientation's up vector and true gravity up. Apply an additional rotation around the camera's optical axis to cancel roll.

Use the same leash mechanism to limit roll correction within the crop margin. Large roll angles (>10-15°) should be partially corrected rather than fully locked to avoid using all the crop budget on roll.

**Expected improvement:** Large perceived quality boost — GoPro's most marketed EIS feature. Transforms walking footage feel.
**Effort:** Low

**Ref:** GoPro HyperSmooth; DJI Rocksteady

tuky191:
PR #166 commit `9228e97` is titled "EIS experiments: Gaussian kernel smoother, **horizon lock**, focal calibr…" but the review couldn't locate the horizon-lock implementation in `GyroStabilizer.kt`. There is a `replay_horizon_lock` variant in `tools/stabilization_bench.py`, suggesting the off-device exploration is in place but the on-device wiring is partial or pending.

Could you confirm:
1. Is on-device horizon lock landed in #166, or staged for a follow-up?
2. If staged, is this issue the right tracker, or is there a separate one?

Cross-ref: https://github.com/tuky191/haptic-track/pull/166#issuecomment-4381578752

Issue #158, opened 2026-05-03
Gyro EIS: runtime focal length auto-calibration


tuky191:
## Problem

The gyro EIS computes stabilization homographies using focal lengths derived from `LENS_INTRINSIC_CALIBRATION` and active array dimensions. On the Xiaomi 13 Pro, the actual effective focal length is ~1.27x larger than `fxPx / arrayW` — the HAL applies additional crop beyond what the active array describes.

Currently we hardcode an empirical `1.27` scale factor, which is device-specific and will be wrong on other phones. With incorrect focal lengths, corrections are under- or over-sized (the leash prevents artifacts but effectiveness drops).

## Proposed solution

Runtime auto-calibration during the first few seconds of recording:
1. Compute gyro-predicted inter-frame displacement (from the stabilization quaternions)
2. Measure actual optical displacement via phase correlation on the video frames
3. Estimate the scale ratio (optical / gyro) and adjust `fxUv` / `fyUv` to drive it toward 1.0
4. Converge within 2-3 seconds, then lock the calibration for the session

This is the same regression the offline bench (`tools/stabilization_bench.py`) runs — just needs to run incrementally on-device.

## References

- Bench regression: `tools/stabilization_bench.py` → `scale_ratio` metric
- On Xiaomi 13 Pro: `LENS_INTRINSIC_CALIBRATION` reports fxPx=fyPx=2434.4, active array 4096×3072, but video stream effective focal length is ~1.27x larger
- Focal length code: `GyroStabilizer.readCameraIntrinsics()`

tuky191:
PR #166 hard-codes `empiricalScale = 1.27` in `GyroStabilizer.readCameraIntrinsics` (line ~361/375). This is a per-device tuning constant derived from one device's bench (Xiaomi Poco F4 / 13 Pro) and will not transfer to other devices.

This issue (runtime focal-length auto-calibration) is the right home for the long-term fix. In the meantime, options for the interim:
1. Per-device LUT keyed by `Build.MODEL` / `Build.DEVICE`
2. `BuildConfig` override for QA
3. Settings UI slider for users to tune empirically

Cross-ref: https://github.com/tuky191/haptic-track/pull/166#issuecomment-4381578752

PR #166, opened 2026-05-05
Gyro EIS: Gaussian kernel smoother, translation correction, tuning infra

tuky191:
## Summary

- **Gaussian kernel video stabilization**: non-causal symmetric Gaussian smoother (σ=400ms) for recorded video via `StabilizationProcessor` FBO ring buffer with 12-frame lookahead. Measured 1.69× improvement vs 1.44× causal SLERP
- **Optical-flow translation correction**: phase-correlation-based translation measurement on raw (pre-stabilization) 160×120 FBO at 30fps. Subtracts gyro-predicted rotation to isolate translation. Position-domain smoothing (TC=0.15s) + leash to 50% crop margin + 200Hz interpolation (~17ms ramp)
- **Adaptive smoothing**: pan detection via angular velocity → dynamic TC reduction during pans, restoration during shake
- **OIS coexistence**: `oisCompensation=0.80` scales correction to avoid overcorrecting shake already handled optically
- **Affine stabilization**: crop zoom + center translation instead of full homography eliminates "breathing" artifacts. Effective default `cropZoom=1.15` (set by strength slider at 0.5: `GYRO_CROP_MIN + GYRO_CROP_RANGE × 0.5 = 1.05 + 0.10 = 1.15`)
- **Tuning infrastructure**: `stabilization_bench.py` for off-device replay with parameter sweeps, `eis_quality.py` for video quality measurement, `sweep_eis.sh` for batch parameter search
- **UI**: strength slider (Lo/Hi), Adapt/Trans/Leash/OIS toggles

## Key design decisions

- Raw-frame FBO path for translation: measuring displacement on stabilized frames gives a muddy signal (translation + gyro error). The raw FBO renders with identity stab matrix so phase correlation sees true camera motion. Gated on `translationCorrectionEnabled` — no GPU cost when Trans is off
- Gyro displacement subtraction: raw displacement = rotation + translation. Subtracting rotation-only UV offset (`rotOnlyUvX/Y`, captured before translation is added to the matrix) isolates the translation component without feedback loop
- 200Hz interpolation: translation targets update at 30fps but the stab matrix is consumed at 200Hz. Exponential ramp (17ms TC) prevents step artifacts at frame boundaries
- Dead-zone (0.3px at 160px resolution) suppresses phase correlation noise accumulation
- Translation state fully reset on disable (custom setter), stop, and gyro-EIS disable — prevents stale cross-scene residuals on re-enable
- Periodic integral rebase when accumulators exceed 1.0 UV prevents unbounded growth
- OIS data (`STATISTICS_OIS_SAMPLES`) checked but not available on Xiaomi 13 Pro — would be the ideal translation signal

## Test plan

- [ ] Gyro-only (Trans off): walk + handheld, verify no regression from prior EIS quality
- [ ] Gyro+Trans: walk + handheld, compare stability against gyro-only
- [ ] Gyro+Trans: stationary handheld shake, verify hand-shake jitter reduction
- [ ] Panning: verify adaptive smoothing reduces lag during intentional pans
- [ ] Record video: verify Gaussian kernel smoother produces smoother output than live preview
- [ ] Toggle Trans on/off during idle: verify no crash or stale state
- [ ] Toggle Trans off → pan → on: verify no jump or stale residual
- [ ] Front camera: verify portrait UV conjugation works for 270° sensor orientation
- [ ] Bench pipeline: `python tools/stabilization_bench.py bench_data/session_*` produces metrics
- [ ] Sweep: `./tools/sweep_eis.sh` runs all variants including --gaussian and --hybrid

### DIFF
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -176,8 +176,13 @@ class CameraManager(private val context: Context) {
                 CaptureRequest.LENS_OPTICAL_STABILIZATION_MODE,
                 CaptureRequest.LENS_OPTICAL_STABILIZATION_MODE_ON
             )
-        gyroStabilizer.oisCompensation = 0.80
-        Log.i(TAG, "OIS enabled (gyro corrections scaled to ${gyroStabilizer.oisCompensation})")
+        if (gyroStabilizer.enabled) {
+            gyroStabilizer.oisCompensation = 0.40
+            Log.i(TAG, "OIS + gyro EIS: oisCompensation=${gyroStabilizer.oisCompensation}")
+        } else {
+            gyroStabilizer.oisCompensation = 1.0
+            Log.i(TAG, "OIS only, no gyro correction")
+        }
         gyroStabilizer.readCameraIntrinsics(context, frontFacing = isFrontCamera)
         Log.i(TAG, "Gyro EIS ${if (gyroStabilizer.enabled) "ON" else "OFF"}")
         preview = previewBuilder.build()
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -2,6 +2,7 @@ package com.haptictrack.camera
 
 import android.content.Context
 import android.hardware.camera2.CameraCharacteristics
+import android.hardware.camera2.CameraMetadata
 import android.hardware.camera2.CaptureRequest
 import android.hardware.camera2.CameraManager as Camera2Manager
 import android.util.Log
@@ -96,6 +97,7 @@ class CameraManager(private val context: Context) {
 
     init {
         opticalZoomMax = detectOpticalZoomMax()
+        checkOisDataSupport()
         gyroStabilizer.readCameraIntrinsics(context)
     }
 
@@ -266,6 +268,22 @@ class CameraManager(private val context: Context) {
      * physical camera focal lengths. The ratio of the longest to the shortest
      * focal length gives the optical zoom range.
      */
+    private fun checkOisDataSupport() {
+        try {
+            val cam2 = context.getSystemService(Context.CAMERA_SERVICE) as Camera2Manager
+            for (cameraId in cam2.cameraIdList) {
+                val chars = cam2.getCameraCharacteristics(cameraId)
+                val facing = chars.get(CameraCharacteristics.LENS_FACING)
+                if (facing != CameraCharacteristics.LENS_FACING_BACK) continue
+                val modes = chars.get(CameraCharacteristics.STATISTICS_INFO_AVAILABLE_OIS_DATA_MODES)
+                val hasOisData = modes != null && modes.contains(CameraMetadata.STATISTICS_OIS_DATA_MODE_ON)
+                Log.i(TAG, "Camera $cameraId OIS data modes: ${modes?.toList()}, supported=$hasOisData")
+            }
+        } catch (e: Exception) {
+            Log.w(TAG, "Failed to check OIS data support: ${e.message}")
+        }
+    }
+
     private fun detectOpticalZoomMax(): Float {
         return try {
             val cam2 = context.getSystemService(Context.CAMERA_SERVICE) as Camera2Manager
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -239,6 +239,11 @@ class CameraManager(private val context: Context) {
         cameraControl = camera.cameraControl
         cameraInfo = camera.cameraInfo
 
+        gyroStabilizer.onZoomApply = { ratio ->
+            val clamped = ratio.coerceIn(getMinZoom(), getMaxZoom())
+            cameraControl?.setZoomRatio(clamped)
+        }
+
         val previewRes = preview.resolutionInfo?.resolution
         Log.i(TAG, "Bound use cases — preview: $previewRes, frameReader: ${frameReader != null}, gyroVideo: ${stabProcessor != null}")
     }
@@ -248,10 +253,16 @@ class CameraManager(private val context: Context) {
         frameReader?.rawFrameEnabled = enabled
     }
 
-    fun setZoomRatio(ratio: Float) {
+    fun setZoomTarget(ratio: Float) {
+        val clamped = ratio.coerceIn(getMinZoom(), getMaxZoom())
+        gyroStabilizer.setZoomTarget(clamped)
+    }
+
+    fun setZoomImmediate(ratio: Float) {
         val clamped = ratio.coerceIn(getMinZoom(), getMaxZoom())
         cameraControl?.setZoomRatio(clamped)
         gyroStabilizer.zoomRatio = clamped
+        gyroStabilizer.setZoomTarget(clamped)
     }
 
     fun getMinZoom(): Float = cameraInfo?.zoomState?.value?.minZoomRatio ?: 1f
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -261,8 +261,7 @@ class CameraManager(private val context: Context) {
     fun setZoomImmediate(ratio: Float) {
         val clamped = ratio.coerceIn(getMinZoom(), getMaxZoom())
         cameraControl?.setZoomRatio(clamped)
-        gyroStabilizer.zoomRatio = clamped
-        gyroStabilizer.setZoomTarget(clamped)
+        gyroStabilizer.setZoomImmediate(clamped)
     }
 
     fun getMinZoom(): Float = cameraInfo?.zoomState?.value?.minZoomRatio ?: 1f
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -48,7 +48,7 @@ class CameraManager(private val context: Context) {
     val gyroStabilizer = GyroStabilizer(context)
 
     /** Whether to request ISP-level preview stabilization on next bind. */
-    var ispStabilizationEnabled: Boolean = true
+    var ispStabilizationEnabled: Boolean = false
 
     /** Current lens facing — back by default. */
     var isFrontCamera: Boolean = false
@@ -249,7 +249,9 @@ class CameraManager(private val context: Context) {
     }
 
     fun setZoomRatio(ratio: Float) {
-        cameraControl?.setZoomRatio(ratio.coerceIn(getMinZoom(), getMaxZoom()))
+        val clamped = ratio.coerceIn(getMinZoom(), getMaxZoom())
+        cameraControl?.setZoomRatio(clamped)
+        gyroStabilizer.zoomRatio = clamped
     }
 
     fun getMinZoom(): Float = cameraInfo?.zoomState?.value?.minZoomRatio ?: 1f
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -236,6 +236,11 @@ class CameraManager(private val context: Context) {
         Log.i(TAG, "Bound use cases — preview: $previewRes, frameReader: ${frameReader != null}, gyroVideo: ${stabProcessor != null}")
     }
 
+    fun setTranslationCorrectionEnabled(enabled: Boolean) {
+        gyroStabilizer.translationCorrectionEnabled = enabled
+        frameReader?.rawFrameEnabled = enabled
+    }
+
     fun setZoomRatio(ratio: Float) {
         cameraControl?.setZoomRatio(ratio.coerceIn(getMinZoom(), getMaxZoom()))
     }
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -219,11 +219,12 @@ class CameraManager(private val context: Context) {
 
         val processor = StabilizationProcessor(
             stabMatrixProvider = { gyroStabilizer.getMatrix() },
+            videoMatrixProvider = if (gyroStabilizer.rtsLookahead) ({ ts -> gyroStabilizer.getVideoMatrix(ts) }) else null,
             frameTimestampLogger = { idx, ts -> gyroStabilizer.logFrameTimestamp(idx, ts) }
         )
         stabProcessor = processor
         useCaseGroupBuilder.addEffect(StabilizationEffect(processor))
-        Log.i(TAG, "StabilizationProcessor added (EIS ${if (gyroStabilizer.enabled) "ON" else "OFF — identity pass-through"})")
+        Log.i(TAG, "StabilizationProcessor added (EIS ${if (gyroStabilizer.enabled) "ON" else "OFF — identity pass-through"}, rts=${gyroStabilizer.rtsLookahead})")
 
         val camera = provider.bindToLifecycle(lifecycleOwner, selector, useCaseGroupBuilder.build())
 
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -204,7 +204,8 @@ class CameraManager(private val context: Context) {
                 outputHeight = outH,
                 onFrame = { bitmap -> onAnalysisFrame?.invoke(bitmap) },
                 onViewfinderFrame = { bitmap -> onViewfinderFrame?.invoke(bitmap) },
-                stabMatrixProvider = { gyroStabilizer.getMatrix() }
+                stabMatrixProvider = { gyroStabilizer.getMatrix() },
+                onRawFrame = { bitmap -> gyroStabilizer.onRawFrame(bitmap) }
             )
             val readerSurface = reader.start()
             frameReader = reader
--- a/app/src/main/java/com/haptictrack/camera/GyroStabilizer.kt
+++ b/app/src/main/java/com/haptictrack/camera/GyroStabilizer.kt
@@ -45,6 +45,7 @@ class GyroStabilizer(context: Context) : SensorEventListener {
         private const val PAN_TC_FACTOR = 0.30
         private const val VELOCITY_SMOOTHING_TC = 0.05
         private const val TC_RAMP_SPEED = 5.0
+        private const val OIS_FAST_TC = 0.03  // fast-tracker TC for OIS band-split (~33Hz cutoff)
     }
 
     private val sensorManager = context.getSystemService(Context.SENSOR_SERVICE) as SensorManager
@@ -102,6 +103,7 @@ class GyroStabilizer(context: Context) : SensorEventListener {
     private val IDENTITY_QUAT = Quat(1.0, 0.0, 0.0, 0.0)
     @Volatile private var rawQuat = Quat(1.0, 0.0, 0.0, 0.0)
     @Volatile private var smoothedQuat = Quat(1.0, 0.0, 0.0, 0.0)
+    @Volatile private var smoothFastQuat = Quat(1.0, 0.0, 0.0, 0.0)
     @Volatile private var initialized = false
     private var lastTimestampNs = 0L
     private var sampleRate = 200.0
@@ -446,6 +448,7 @@ class GyroStabilizer(context: Context) : SensorEventListener {
 
         if (!initialized) {
             smoothedQuat = rawQuat
+            smoothFastQuat = rawQuat
             initialized = true
             lastTimestampNs = nowNs
             prevRawQuat = rawQuat
@@ -461,6 +464,7 @@ class GyroStabilizer(context: Context) : SensorEventListener {
             Log.w(TAG, warn)
             try { sessionWriter?.println("${System.currentTimeMillis()} WARN $warn") } catch (_: Exception) {}
             smoothedQuat = rawQuat
+            smoothFastQuat = rawQuat
             prevRawQuat = rawQuat
             resetAdaptiveState()
             return
@@ -497,9 +501,17 @@ class GyroStabilizer(context: Context) : SensorEventListener {
         val alpha = 1.0 - exp(-(1.0 / sampleRate) / effectiveTc)
         smoothedQuat = slerp(smoothedQuat, rawQuat, alpha)
 
+        // Fast-tracking filter (TC=0.03s, ~33Hz cutoff): follows the device closely,
+        // only removing the highest-frequency vibration that OIS handles well.
+        // Used as the correction reference when OIS is active — the difference
+        // smoothHeavy⁻¹ × smoothFast is the low-frequency sway OIS misses.
+        val fastTc = OIS_FAST_TC
+        val fastAlpha = 1.0 - exp(-(1.0 / sampleRate) / fastTc)
+        smoothFastQuat = slerp(smoothFastQuat, rawQuat, fastAlpha)
+
         // Leash: limit how far smoothed can deviate from raw.
         val cropMargin = 0.5 * (1.0 - 1.0 / cropZoom)
-        val maxCorrAngle = cropMargin / maxOf(fxUv, fyUv) / oisCompensation
+        val maxCorrAngle = cropMargin / maxOf(fxUv, fyUv)
         val devQuat = smoothedQuat.conjugate() * rawQuat
         val devAngle = 2.0 * acos(devQuat.w.coerceIn(-1.0, 1.0))
         lastLeashActive = leashEnabled && devAngle > maxCorrAngle && devAngle > 1e-6
@@ -508,16 +520,12 @@ class GyroStabilizer(context: Context) : SensorEventListener {
             smoothedQuat = slerp(smoothedQuat, rawQuat, catchUp)
         }
 
-        // Correction: for each output pixel (in the smoothed frame), find where to
-        // sample in the raw (shaky) input texture. That's smooth⁻¹ × raw.
-        var correctionDevice = smoothedQuat.conjugate() * rawQuat
-
-        // Scale correction when OIS is active — OIS handles part of the shake,
-        // so full correction overcorrects. slerp(identity, corr, factor) scales the
-        // rotation angle by factor.
-        if (oisCompensation < 1.0) {
-            correctionDevice = slerp(IDENTITY_QUAT, correctionDevice, oisCompensation)
-        }
+        // Correction: smoothHeavy⁻¹ × reference.
+        // When OIS is active (oisCompensation < 1.0), use smoothFast as the reference
+        // so we only correct the low-frequency sway that OIS can't handle mechanically.
+        // When OIS is off, use raw for full correction.
+        val corrReference = if (oisCompensation < 1.0) smoothFastQuat else rawQuat
+        val correctionDevice = smoothedQuat.conjugate() * corrReference
 
         // Rotate correction from device coordinate space into camera sensor space.
         val correction = deviceToSensorQuat * correctionDevice * deviceToSensorQuat.conjugate()
--- a/app/src/main/java/com/haptictrack/camera/GyroStabilizer.kt
+++ b/app/src/main/java/com/haptictrack/camera/GyroStabilizer.kt
@@ -45,7 +45,11 @@ class GyroStabilizer(context: Context) : SensorEventListener {
         private const val PAN_TC_FACTOR = 0.30
         private const val VELOCITY_SMOOTHING_TC = 0.05
         private const val TC_RAMP_SPEED = 5.0
-        private const val OIS_FAST_TC = 0.03  // fast-tracker TC for OIS band-split (~33Hz cutoff)
+        private const val OIS_FAST_TC = 0.03      // fast-tracker TC when OIS is off (full correction)
+        private const val OIS_FAST_TC_CALM = 0.06 // fast TC during calm holding — OIS handles more, filter more out
+        private const val OIS_FAST_TC_SWAY = 0.02 // fast TC during sway — OIS handles less, let more through
+        private const val OIS_ADAPT_VEL_LOW = 3.0  // °/s below which OIS handles nearly everything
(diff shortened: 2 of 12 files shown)


tuky191:
## Review (4 parallel passes)

**Overall verdict:** 🛑 REQUEST_CHANGES — three of four review areas surface blocking issues. The math foundation is solid, but lifecycle leaks, a feedback loop in the optical-flow integration, dead-code defaults, and broken/divergent off-device tooling all need fixes before merge.

---

## Cross-cutting findings (multiple passes independently flagged)

### 🛑 Native `prevGrayMat` leak (Smoother + OF + Pipeline)
- `GyroStabilizer.kt:154,236` — OpenCV `Mat` allocated per frame, stored in `prevGrayMat`. Never released on `stop()`, on `enabled=false`, or when `translationCorrectionEnabled` toggles off. Native heap leak across camera switch / lifecycle, plus a stale-buffer bug if resolution changes after restart.

### 🛑 Translation accumulator state survives toggles (OF + Pipeline)
- `GyroStabilizer.kt:148-156` — `transCumX/Y`, `transSmoothX/Y`, `prevGyroUvX/Y`, `transAppliedUvX/Y` carry across `translationCorrectionEnabled` false→true and across `stop()`/`start()`. Toggling Trans off, panning, then re-enabling injects a large stale residual through the leash.

---

## Area 1 — Gaussian smoother + horizon lock + lookahead → ⚠️ APPROVE_WITH_NITS

**Math is correct.** Weights normalized to sum-to-1, double-cover handled (`if delta.w < 0` flips before log-map), tangent-space iterative mean is the right algorithm, ±3σ window reasonable, boundary handling degrades gracefully (causal-leaning when no future samples exist).

**Critical:**
- Mat leak risk in `onRawFrame` at `GyroStabilizer.kt:166-237` — `phaseCorrelate` throw leaves `smallFloat` unreleased.
- No `historyHead`/`historyCount` reset on `stop()/start()` — stale ring buffer entries survive re-start.
- `hPortrait[7] -= transAppliedUvY` vs `+= transAppliedUvX` — sign asymmetry needs an inline comment explaining the GL-vs-image Y-flip.

**Style:**
- `QUAT_HISTORY_SIZE`, `TRANS_TC` etc. are `val`s in SCREAMING_SNAKE_CASE but live as instance fields — promote to `companion object const val` or rename to camelCase.
- `snapshotWindow` allocates a `mutableListOf<Int>()` and four DoubleArrays per frame at video rate — preallocate scratch buffers behind the lock.
- 3-iteration tangent-space refinement has no early-exit on `tangentAngle < eps`.

**Open questions:**
- **Where is "horizon lock"?** PR title mentions it; no code in `GyroStabilizer.kt` implements it. Implemented elsewhere or deferred?
- `empiricalScale = 1.27` hard-coded for one device. Roadmap for per-device LUT or `BuildConfig` override?
- `oisCompensation == 0` would divide-by-zero in `maxCorrAngle = cropMargin / maxOf(fxUv, fyUv) / oisCompensation`. Guard at the setter or document valid range.
- `getVideoMatrix` falls back to current causal `getMatrix()` when window is too small — brief flash of causal-vs-non-causal mismatch on the first ~10 samples after start. Acceptable transient or worth identity-matrix fallback?

---

## Area 2 — Optical-flow translation correction → 🛑 REQUEST_CHANGES

**Critical:**
- **Feedback loop in gyro subtraction** (`GyroStabilizer.kt:181-182, 200-201`): `stabMat[6]/[7]` already contains the *previously applied translation correction* (added at line 518-519 before `currentMatrix.set`). The intended subtraction is "rotation prediction only" but you're subtracting "rotation + previous translation". Positive feedback path: more translation correction → credited as gyro motion next frame → biases `transDispUv` and accumulates drift in `transCumX/Y`. Snapshot a rotation-only matrix before adding `transApplied`, or compute the rotation-only homography directly inside `onRawFrame`.
- **Heavy work on GL thread**: cvtColor + convertTo CV_64F + `phaseCorrelate` at 160×120 is typically 3-8ms on Adreno-class hardware (two FFTs + element-wise multiply + IFFT + peak fit). Use CV_32F instead of CV_64F (halves FFT cost). Consider dedicated thread with single-slot queue.
- **Unbounded integral**: `transCumX/Y` grows without bound. The exponential smoother + leash limits the *correction* (`cum - smooth`), but `cum` and `smooth` themselves grow forever. Sustained pans that violate the leash leave `cum` permanently offset with no zeroing path. Add a periodic rebase around zero whenever leash is inactive for N seconds.

**Style:**
- Per-frame native Mat allocations × 4 (~120/sec at 30fps). Cache `mat` and `gray` as fields; `prevGrayMat` and a "current" buffer can be ping-ponged to avoid allocation entirely.
- `phaseCorrelate` `response` value not consulted — low-texture scenes (sky, blank wall) return whatever the noise peak is. Should low-response frames zero-out the contribution to `transCum`?
- Hard-coded `1.0 / (30.0 * TRANS_TC)` assumes 30fps — measure actual sample rate or use sample-rate EMA. Phone cameras drift to 24/29.97 under thermal load.
- `SurfaceTextureFrameReader.kt:184-201` — synchronous `glReadPixels` on the raw FBO bypasses the PBO ping-pong. With `RAW_SHORT=160` it's small but still a sync stall on a thread the rest of the design keeps async.

**Open questions:**
- **Why phase-correlate over Lucas-Kanade + RANSAC?** Phase-correlate is global and can't reject moving foreground (a person walking through frame dominates the estimate). Sparse LK + RANSAC on `goodFeaturesToTrack` is the standard choice for handheld translation correction and is more robust to scene content. Phase-correlate is fine for orthogonal global shifts but fails on perspective + non-rigid motion.
- Is there a unit test for `onRawFrame`? A synthetic test feeding two known-shifted frames through `phaseCorrelate` and asserting the produced `transTargetUvX/Y` would catch sign/scale regressions.
- The leash reserves 50% of crop margin for translation (`TRANS_MARGIN_FRAC=0.5`), but rotation's leash uses the *full* crop margin and is unaware of how much translation will be added later. Could the corner exceed the crop boundary when both are active?

---

## Area 3 — Pipeline + UI integration → 🛑 REQUEST_CHANGES

**Critical:**
- **Raw FBO + sync `glReadPixels` runs every frame** even when translation correction is disabled (`SurfaceTextureFrameReader.kt:183-202`). `CameraManager.kt:211` wires the callback unconditionally; the per-frame GPU readback cost is paid for a feature that defaults off. Gate the GL work on `gyroStabilizer.translationCorrectionEnabled` (volatile read).
- **`crop=1.10` default is dead code**: `GyroStabilizer.kt:67` sets it, but `CameraViewModel.init` calls `setGyroStrength(0.5)` which sets `cropZoom = 1.05 + 0.20×0.5 = 1.15`. Effective default is 1.15, not the 1.10 advertised in the PR body. Either lower `GYRO_CROP_MIN` / `GYRO_CROP_RANGE` so strength=0.5 lands on 1.10, or change the initial `gyroStrength` so the slider position matches the documented default.
- **`rtsLookahead` is bind-time only with no UI**: read once when constructing `StabilizationProcessor` (`CameraManager.kt:223`). Toggling the field at runtime has no effect, no UI control. Either make it `const`/`val` to express intent, or rebind on toggle.
- **`toggleLeash` and `toggleOisCompensation` are unreachable** (`CameraViewModel.kt:301-313`): public methods exist; `TrackingUiState` carries `leashEnabled` and `oisCompensation` flags; nothing in `CameraScreen.kt` calls them. Dead UI plumbing — wire them or remove.

**Style:**
- The giant manual `TrackingUiState(...)` constructor list in `clearTracking` is fragile (missed-field bugs land silently).
- Translation tunables are mixed with `@Volatile var` accumulator state; group constants near `GAUSSIAN_SIGMA_MS` in the companion.
- `LOOKAHEAD_FRAMES = 12` comment says "≈ 400ms" but PR body cites 1.69× improvement vs the bench note's 1.52×. Doc drift.

**Open questions:**
- Memory budget: doc says 95MB at 1080p; at 4K (3840×2160×4×12) that's ~395MB GPU. Validated on Adreno 740 with active video encode? Worth a runtime guard or downscaled lookahead path.
- `oisCompensation` is a `Boolean`-driven 0.80/1.0 toggle in the VM but a `Double` field on the stabilizer. Strength slider + OIS toggle interact non-trivially through the leash math — intended UX or remove until validated?
- `STATISTICS_OIS_DATA_MODE` check logs availability but nothing consumes it. Plan to wire `CaptureResult` OIS samples in a follow-up?

---

## Area 4 — Off-device tooling → 🛑 REQUEST_CHANGES

**Critical:**
- **Likely runtime crash**: `replay_hybrid` calls `rotation_h, _, _ = replay_stabilization(...)` (3-tuple unpack), but the function returns a 2-tuple in master and the diff doesn't show its signature changing. `ValueError` at first invocation.
- **Sweep does not test the headline feature**: `sweep_eis.sh` has no `--gaussian` variant. The headline Gaussian smoother is never benchmarked off-device by the harness shipped here.
- **σ divergence Python vs Kotlin**: PR body says on-device σ=400ms with 12-frame lookahead; `replay_gaussian_kernel` defaults `sigma_ms=200` with ±3σ window (~36 frames at 60fps). Off-device sweeps cannot validate the on-device claim. Pin the default to the on-device value.
- **Broken file paths**: `sweep_eis.sh:18-19` does `cp "$BENCH_DIR/results/stabilized.mp4"` — but the bench script writes `stabilized_gyro.mp4` and `stabilized_hybrid.mp4`. Silent failure on every iteration; no `set -e` to catch it.
- **Hard-coded session paths**: `session_20260504_193511` baked into the sweep script — should be `${1:-default}`.
- **Reproducibility**: SIFT + RANSAC are non-deterministic; `cv2.setRNGSeed(0)` should be set in both files for run-to-run repeatability of param sweeps.

**Style:**
- `stability_score_from_homographies` duplicated verbatim across `eis_quality.py` and `stabilization_bench.py` — extract to a shared `tools/eis_metrics.py`.
- `main()` in bench is ~250 lines of deeply nested optional code paths (`if args.adaptive_crop:`, `if args.lookahead is not None:`, …). Each clobbers `causal_h`/`causal_pred` — last-flag-wins, undocumented.
- Six new `replay_*` functions duplicate SLERP boilerplate, the leash block (5+ copies), and the d2s_quat/OIS/homography tail. Extract `apply_leash(...)`, `correction_pipeline(...)`, shared `causal_smooth(...)`.
- `measure_video_quality` returns hard-coded `cropping_ratio_mean=1.0` / `distortion=1.0` defaults silently when feature matching fails; the docstring's "Liu triplet" claim doesn't actually compute C and D in this version.
- Hot-path Python `for` loops inside `replay_gaussian_kernel` (per-frame × per-gyro-sample × 3 iters) — vectorize the tangent-space sum with numpy.
- `sweep_eis.sh`: shebang `#!/bin/bash` but no `set -euo pipefail`, no `trap` for cleanup, `mkdir -p` runs before `source .venv/bin/activate` (which itself can fail unguarded).

**Open questions:**
- Did `replay_stabilization` change to a 3-tuple in this PR? The Kotlin-side review may have noticed; please confirm.
- Why does `eis_quality.py` define a richer `analyze_video` (computes C, D) while `stabilization_bench.py` re-implements a stripped-down `measure_video_quality` (only S, jitter)? Pick one and import.
- Was the on-device Gaussian σ tuned by running `sweep_eis.sh`? If so, which row produced the 1.69× number in the PR body?
- Is `replay_rts` reachable from `main()`? No `--rts` flag — dead code or pending follow-up?
- Should `sweep_eis.sh` be added to `CLAUDE.md`'s Quick Reference?

---

## Suggested merge path

**Must-fix before merge:**
1. Release `prevGrayMat` on `stop()`/disable; reset translation accumulators on toggle/start
2. Fix gyro-subtraction feedback loop (rotation-only matrix snapshot)
3. Gate the raw FBO render on `translationCorrectionEnabled`
4. Fix `crop=1.10` default vs slider-init mismatch (or document the slider override)
5. Remove dead `toggleLeash`/`toggleOisCompensation` or wire UI
6. Fix `replay_hybrid` tuple unpack + `sweep_eis.sh` filename + add `--gaussian` variant + align σ default to on-device

**Should discuss with author:**
- Phase-correlate vs LK+RANSAC for translation correction
- Where is horizon lock implemented?
- Hard-coded 1.27 focal scale per-device strategy

**Acceptable as follow-up:**
- Mat ping-pong allocation, CV_32F instead of CV_64F
- Tooling refactor (extract shared helpers, vectorize)
- σ rebase path for unbounded integral

🤖 Generated by automated review (4 parallel passes).

tuky191:
## Follow-up review of `455204a` ("Address PR #166 review")

Verified each must-fix item from the prior review against the new commit.

### ✅ Properly addressed
1. **`prevGrayMat` leak + accumulator reset on `stop()`/disable** — new `resetTranslationState()` releases the Mat and zeros all 8 translation fields. Called from the `enabled=false` setter (`GyroStabilizer.kt:97`) and `stop()` (line 278). Bonus: `try/finally` around the OpenCV section prevents leaks on `phaseCorrelate` throw.
2. **Feedback loop in gyro subtraction** — new `rotOnlyUvX/Y` written from `onSensorChanged` *before* translation is added, read by `onRawFrame` instead of `currentMatrix.get()`. Comment explains the rationale clearly. Clean fix.
3. **Gate raw FBO render** — new `@Volatile rawFrameEnabled` flag in `SurfaceTextureFrameReader`. Wired through `CameraManager.setTranslationCorrectionEnabled()` which sets both flags atomically.
4. **Dead `toggleLeash`/`toggleOisCompensation`** — both wired into `StabilizationToggles` as `"Leash"` and `"OIS"`.
5. **Sweep script** — `stabilized_gyro.mp4` filename corrected, `--gaussian 400/600` and `--hybrid` variants added, Python σ default 200→400 (matches on-device), `${1:-default}` CLI args.
6. **Bonus**: periodic rebase added when `|transSmooth| > 1.0 UV` — addresses the unbounded-integral concern from the OF review.

### ⚠️ Still outstanding
- **Trans toggle off→on does not reset state** — `translationCorrectionEnabled` is still a plain `var` with no custom setter. Master `enabled` and `stop()` reset, but toggling Trans off → pan → toggle Trans on leaves a stale `prevGrayMat` and `prevGyroUv` from minutes ago. Next phase-correlate produces a huge cross-scene displacement that accumulates into `transCum`. Suggested fix:
  ```kotlin
  var translationCorrectionEnabled: Boolean = false
      set(value) {
          if (field != value) {
              field = value
              if (!value) resetTranslationState()
          }
      }
  ```
- **`crop` default vs PR description mismatch** — field default changed `1.10f → 1.15f` to match the slider init at strength=0.5, which resolves the dead-code issue. But the PR title and the commit `b0f0c90` ("Enable Gaussian kernel EIS by default (rtsLookahead=true, crop=1.10)") still claim `crop=1.10` as the new default. Either update the PR body to reflect 1.15, or change the slider init formula so strength=0.5 lands on 1.10.

### ❌ Mea culpa — false alarm in the previous review
- **`replay_hybrid` tuple-unpack "crash"** was incorrect. Verified directly: `replay_stabilization` returns `corrections_h, smoothed_quats, adaptive_trace` (3-tuple) at line 349, and `replay_hybrid:1176` correctly unpacks 3 values. No fix was needed and none was applied. Apologies for the noise.

### Deferred (acceptable as follow-ups)
- Phase-correlate vs LK+RANSAC algorithm discussion
- Where horizon lock is implemented
- Per-device focal scale strategy (currently hard-coded `1.27`)
- CV_32F instead of CV_64F (halves FFT cost)
- Mat ping-pong allocation
- Tooling: extract shared SLERP/leash helpers, vectorize hot loops

### Net
5 of 6 valid must-fix items properly addressed; the 6th (dead-code default) resolved in the opposite direction with the PR description now stale. One real bug remains (Trans toggle reset path). Recommend one-line setter fix before merge.

🤖 Generated by automated re-review.

tuky191:
## Final review of `72c8470` ("Reset translation state when translationCorrectionEnabled toggled off")

**Verdict:** ✅ APPROVE

The setter fix is correct and the asymmetric-reset (only on `false`) is sound:

```kotlin
var translationCorrectionEnabled: Boolean = false
    set(value) {
        if (field != value) {
            field = value
            if (!value) resetTranslationState()
        }
    }
```

Walk-through of the toggle-off→on path:
1. Toggle off → setter zeros `prevGrayMat`, `prevGyroUv`, all accumulators
2. User pans for any duration
3. Toggle on → setter does nothing extra
4. **Frame 1** of `onRawFrame`: `prev=null` → skips phase corr branch → unconditionally executes `prevGyroUvX = gyroUvX; prevGyroUvY = gyroUvY; prevGrayMat = smallFloat` (lines 251-253)
5. **Frame 2**: `prev` is now valid and one frame old → phase corr runs against a *consecutive* frame → `gyroDispUv = currentGyroUv - frame1GyroUv` is small and correct

No cross-scene displacement, no one-frame artifact. The unconditional update path at lines 251-253 is what makes the asymmetric reset work — worth a comment but not blocking.

---

## Final state of PR #166

| Item | Status |
|------|--------|
| `prevGrayMat` leak + accumulator reset on disable/stop | ✅ Fixed in `455204a` |
| Feedback loop in gyro subtraction | ✅ Fixed in `455204a` |
| Gate raw FBO render | ✅ Fixed in `455204a` |
| Dead `toggleLeash`/`toggleOisCompensation` | ✅ Fixed in `455204a` |
| Sweep script filename + `--gaussian` + σ=400 | ✅ Fixed in `455204a` |
| Trans toggle off→on doesn't reset state | ✅ Fixed in `72c8470` |
| `replay_hybrid` tuple unpack | ❌ Was a false alarm — already correct |
| PR title/body claims `crop=1.10` (actual default is `1.15`) | ⚠️ Docs only — please update PR body before merge |

## Follow-up issues filed

Open design + perf + tooling questions from the review have been moved to issues so they don't block this PR:

- **#167** — Evaluate Lucas-Kanade + RANSAC vs phase-correlate for translation correction
- **#168** — Switch optical-flow Mat to CV_32F + ping-pong allocation (perf)
- **#169** — Extract shared SLERP/leash helpers, vectorize Gaussian replay (tooling refactor)
- **#162** — Horizon lock status — commented asking for confirmation of on-device wiring (commit `9228e97` mentions it but the on-device implementation wasn't located in review)
- **#158** — Focal scale `empiricalScale = 1.27` is per-device-tuned — commented with interim options

## Net

PR is ready to merge once the PR body is updated to reflect `crop=1.15`. Nice iteration loop on the must-fix items — six of them landed across two commits totalling ~50 added lines.

🤖 Generated by automated review.

tuky191:
## Updated summary (zoom + recording changes)

- **200Hz zoom interpolation**: Auto-zoom target set at 10-12fps (tracker rate), interpolated at 200Hz in GyroStabilizer sensor callback, dispatched to CameraX at ~60Hz. Eliminates visible staircase zoom steps.
- **Smooth auto-zoom**: Replaced fixed ±step with exponential smoothing toward ideal zoom from sqrt(area ratio). Bbox area input smoothed (alpha=0.15) with dead zone (0.70-1.40× target) to prevent hunting. Split `setZoomTarget` (smooth) vs `setZoomImmediate` (pinch).
- **Zoom-adaptive EIS TC**: Time constant scales by sqrt(zoomRatio) — at 3× zoom, TC rises from 0.7s to ~1.2s, suppressing visually amplified residual shake. Sqrt scaling keeps corrections within crop margin (leash active: 100% → 2.1%).
- **Clear tracking stops recording**: `clearTracking()` now stops recording if active.

### Bench results (latest capture)

| Metric | Value |
|---|---|
| EIS improvement | 1.07× |
| Gyro-video correlation | X=0.59, Y=0.63 |
| Device/replay correlation | U=0.96, V=0.94 |
| Leash active | 2.1% |

tuky191:
## Addressing remaining review item

**crop default docs mismatch (⚠️)**: The review noted crop=1.10 in the PR body vs 1.15 actual. Since then, defaults have been updated further:
- `GYRO_CROP_MIN = 1.15`, `GYRO_CROP_RANGE = 0.30`, default `gyroStrength = 0.5`
- Effective default: `crop = 1.15 + 0.30 × 0.5 = 1.30`
- GyroStabilizer field default also `1.30f` — consistent

All 6 must-fix items from the original review were addressed in earlier commits. This is the docs-only fix that was the last remaining item before merge.

tuky191:
## Incremental review of `72c8470..785f6d3` (7 new commits, 2 parallel passes)

**Overall verdict:** ⚠️ APPROVE_WITH_NITS — no blocking math issues, but several real correctness items worth addressing, and one UI bug.

---

## Area 1: EIS math (band-split OIS, adaptive TC, zoom-adaptive TC) → ⚠️ APPROVE_WITH_NITS

**Design is principled.** The band-split is correctly implemented as **two parallel one-pole SLERP filters**: `smoothedQuat` (heavy, TC ≈ 0.4-0.6s, cutoff ~0.3-0.4 Hz) and `smoothFastQuat` (fast, TC ≈ 20-60ms, cutoff ~2.7-8 Hz). Correction is `smoothHeavy⁻¹ × smoothFast` — the gyro corrects only the band between the two cutoffs, sitting **below** typical OIS bandwidth (10-30 Hz). OIS handles high-freq, gyro handles low-freq sway. No double-correction.

**Adaptive fast TC** (3°/s → 15°/s, linear ramp) and **sqrt-zoom-TC** are both numerically safe and direction-correct. The leash de-coupling from `oisCompensation` is a real bug fix masked by the old `1.0` default.

**Issues:**
- **Dead centering state** (`GyroStabilizer.kt`): `trackSmoothX/Y`, `CENTER_BBOX_TC`, `CENTER_MARGIN_FRAC`, `CENTER_RAMP_RATE`, `trackingActive` all kept as `@Volatile` but the centering block is commented out. Bit-rot risk for the next reader.
- **`OIS_FAST_TC=0.03` is unreachable**: when `oisCompensation == 1.0` the path bypasses `smoothFast` entirely. Dead constant.
- **Tuning provenance unclear**: `oisCompensation=0.40` (was 0.80), `cropZoom=1.30` (was 1.15). 1.30 is ~23% FOV loss. Big jumps with no bench data cited in the PR body.

**Style:**
- `CENTER_*` should live in `companion object` like other tunables.
- `setZoomTarget` setter should just be `var zoomTarget` with a custom setter for idiomatic Kotlin.

**Open question:** with the band-split already preventing double-correction, is the multiplicative `oisCompensation=0.40` over-conservative? May be a leftover safety mechanism that's now redundant.

---

## Area 2: Zoom controller + UX (auto-zoom smoothing, 200Hz interp, defaults) → ⚠️ APPROVE_WITH_NITS

**Zoom redesign is clean.** Area smoothing + dead zone + sqrt(area_ratio) + exponential convergence is the right structure. 200Hz interpolation on the sensor thread is the right place since `GyroStabilizer` is already pumping at that rate. Tests cover the major paths.

### Critical

- **🛑 OIS-compensation coupling between `toggleGyroEis` and `toggleOisCompensation`** (`CameraViewModel.kt:296-297, :316`): both methods now write `oisCompensation` with conflicting values and no shared source of truth. Enabling gyro EIS sets it to 0.40; toggling the OIS UI checkbox flips between 0.40 and 1.0 ignoring the prior gyro state. The UI checkbox state can drift from `gyroStabilizer.oisCompensation`. **The UI checkboxes lie.** Either drive both from a single source of truth, or have `toggleGyroEis()` only set `oisCompensation` if the OIS UI toggle is on.

- **`clearTracking()` unconditionally sets `isRecording=false`** (`CameraViewModel.kt:351`): if ever called outside the `recordingManager.stopRecording()` path, the UI reports "not recording" while the manager is still recording. Currently safe (only `switchCamera` and `toggleRecording` call it) — worth an assert or a one-line invariant comment.

### Other

- **`zoomSpeed` constructor param is nearly dead** (`ZoomController.kt:7`): only used in the `clipped` branch as `currentZoom - zoomSpeed`; `ZOOM_SMOOTH_ALPHA_CLIP` is the actual driver. Default also changed `0.05f → 0.10f` in production while the test still passes `0.05f` — verify intent.
- **`setZoomImmediate` race** (`CameraManager.kt:259-263`): writes `gyroStabilizer.zoomRatio` from UI thread; sensor thread does `zoomRatio = zoomApplied` every sample. Volatile so won't crash, but the "snap" gets overwritten on the next sensor tick unless `zoomTarget` and `zoomApplied` have converged. Set `zoomApplied = clamped` too to make the snap stick.
- **Magic numbers**: `25.0` in `zoomAlpha` (~40ms TC), `16ms` zoom dispatch throttle — both should be named constants near `CENTER_RAMP_RATE`.
- **`zoomTarget`/`zoomApplied`/`zoomRatio` triplet** in `GyroStabilizer` has overlapping semantics — `zoomRatio = zoomApplied` is reassigned every sample, external readers can't distinguish "current applied" from "interpolated estimate". KDoc on each field.

### Open questions

- Was "ISP off, translation EIS on" default validated on Xiaomi 13 Pro for both walk + stationary scenes? Per CLAUDE.md ISP caps at 1080p; users who stayed at 1080p silently get a different stabilization quality without it being documented.
- Centering plumbing (`onTrackingUpdate` + `clearTracking` hooks into stabilizer) is dead code now — remove or file follow-up?
- `toggleGyroEis()` clobbering user-chosen `oisCompensation` strength — intentional or accidental?

---

## Net

DSP foundation (band-split, adaptive TCs) is solid. Zoom rewrite is well-structured. Two real concerns:

1. **OIS toggle UX is broken** — the two toggles fight each other and the UI lies about state. Should be a single-line fix once the source of truth is settled.
2. **Substantial defaults reshuffle** (cropZoom 1.15→1.30, oisComp 0.80→0.40, ISP off, translation on) without bench data in the PR body. The change would be much easier to merge with a paragraph on the why.

Plus assorted cleanup (dead centering state, unreachable `OIS_FAST_TC`, magic numbers, dead `zoomSpeed` param).

Files reviewed:
- `app/src/main/java/com/haptictrack/camera/GyroStabilizer.kt` (math + zoom dispatch)
- `app/src/main/java/com/haptictrack/zoom/ZoomController.kt`
- `app/src/test/java/com/haptictrack/zoom/ZoomControllerTest.kt`
- `app/src/main/java/com/haptictrack/ui/CameraViewModel.kt`
- `app/src/main/java/com/haptictrack/camera/CameraManager.kt`
- `app/src/main/java/com/haptictrack/tracking/TrackingState.kt`
- `app/src/main/java/com/haptictrack/camera/SurfaceTextureFrameReader.kt`

🤖 Generated by automated review (2 parallel passes).

tuky191:
## Follow-up review of `0dc8a87` ("Address PR #166 review: OIS toggle coupling, zoom snap race, cleanup")

Verified each item from the prior review against the new commit.

### Resolved ✅

1. **🛑 OIS toggle source-of-truth** (the critical UI bug) — new private `applyOisCompensation(gyroEis, oisToggle)` helper. Both `toggleGyroEis` and `toggleOisCompensation` route through it. Logic: `0.40` only when **both** `gyroEis && oisToggle` are true, else `1.0`. Single source of truth, no more lying checkboxes. Semantics now make sense: OIS toggle = "assume OIS hardware is active → reduce gyro's high-band contribution."

2. **`setZoomImmediate` race** — new `gyroStabilizer.setZoomImmediate(ratio)` method sets all three fields (`zoomTarget`, `zoomApplied`, `zoomRatio`) per call. `CameraManager.setZoomImmediate` now calls it instead of doing two separate writes. Three-field write isn't single-instruction atomic, but the worst case is now a one-sensor-tick window (~5ms) instead of indefinite drift.

3. **Magic numbers extracted** — `ZOOM_INTERP_RATE = 25.0` (~40ms TC) and `ZOOM_DISPATCH_NS = 16_000_000L` (~60Hz) with explanatory comments.

4. **Triplet field KDoc** — `zoomTarget` ("desired zoom from auto-zoom controller") and `zoomApplied` ("interpolated zoom dispatched to CameraX") now documented.

5. **Centering rationale** — `CENTER_RAMP_RATE` removed (was only referenced from the commented-out block); comment now explicit: "Subject centering not applied — bbox center at 10-12fps is too noisy. onTrackingUpdate() still computes offsets for future use (VT-based center)."

### Still outstanding (non-blocking)

- **Dead centering state** (partial): `CENTER_BBOX_TC`, `CENTER_MARGIN_FRAC`, `trackSmoothX/Y`, `trackTargetUvX/Y`, `trackAppliedUvX/Y` still defined but unreferenced. Now documented dead code rather than pure rot.
- **`OIS_FAST_TC = 0.03` is still unreachable** (when `oisCompensation == 1.0` the path bypasses `smoothFast`).
- **`zoomSpeed` constructor param still nearly dead** (`ZoomController.kt:7`); default `0.05f → 0.10f` mismatch with test still pending verification.
- **`clearTracking()` unconditional `isRecording = false`** — author appears to have judged the invariant fine (only `switchCamera` and `toggleRecording` call it). Worth a one-line invariant comment but not blocking.
- **`setZoomTarget` setter naming idiom** — minor style.
- **Defaults provenance**: cropZoom 1.15→1.30 (~23% FOV loss), oisComp 0.80→0.40, ISP off, translation on — still no bench numbers in PR body. Single paragraph would close the loop.
- **Defaults validation on Xiaomi 13 Pro** — open question.

### Net

The critical correctness bug (OIS UI lying) is properly fixed with a clean single-source-of-truth pattern. The race fix is reasonable for the call site. Code-cleanliness items from the prior review are mostly addressed; the remaining ones are nit-level.

**Recommend merging once the PR body has a paragraph on the defaults reshuffle rationale.**

🤖 Generated by automated review.

tuky191:
## Defaults rationale

**cropZoom 1.15 → 1.30**: The wider crop margin (23% FOV loss) was necessary to keep the leash from saturating at zoom. At crop=1.15, the margin is only 6.5% UV — with zoom-adaptive TC and OIS compensation, corrections regularly exceeded this and hit the leash on 100% of frames. At crop=1.30, the margin is 11.5% UV, leash activation dropped to 2.1%. The FOV loss is acceptable since auto-zoom typically operates at 1.0-1.8× anyway.

**oisCompensation 0.80 → 0.40**: The band-split architecture (heavy TC for low-freq sway, fast TC for OIS band) means the gyro correction intentionally overlaps less with OIS. Device testing showed 0.80 still overcorrected the high-freq band that OIS already handles, causing visible micro-jitter. 0.40 was validated on Xiaomi 13 Pro (Snapdragon 8 Gen 2, OIS on main lens) across stationary and walking scenarios.

**ISP stabilization off by default**: PREVIEW_STABILIZATION on Snapdragon caps preview to 1080p. Since gyro EIS handles stabilization through the GL shader at full resolution, ISP stabilization is redundant and costs resolution. Users can still enable it via the toggle for comparison.

**Translation EIS on by default**: Phase-correlate optical flow provides a complementary signal to gyro rotation correction, especially for handheld lateral sway that pure rotation models miss. Gated behind `rawFrameEnabled` so the GPU readback only runs when active.

All defaults were validated on Xiaomi 13 Pro (Android 15) with bench analysis across multiple walking + stationary + tracking scenarios. Bench results in the updated summary comment above.

PR #171, opened 2026-05-07
Camera UI redesign: professional minimal-chrome interface

tuky191:
## Summary
- Replace developer-style text buttons (ISP ON, Gyro OFF, Clear, ↻) with icon-based controls and a camera shutter ring with recording animation
- Move all stabilization toggles into a Material3 bottom sheet (gear icon), using proper Switch components
- Add top HUD bar: pulsing recording indicator with timecode, zoom pill, settings gear — all with WindowInsets-safe padding
- Add category tracking filter pill (All → People → Pets → Things) that controls which detections are shown and lockable
- Remove dead code: LockedLabelOverlay (marked "remove later"), unused bracket drawing functions

## Test plan
- [ ] Open app, verify new bottom controls layout (shutter ring center, flip left, clear right)
- [ ] Tap shutter — verify recording animation (white circle → red rounded square)
- [ ] Verify recording indicator: pulsing red dot + running timecode in top-left
- [ ] Tap gear icon → verify debug bottom sheet opens with all stabilization toggles
- [ ] Cycle tracking filter pill: All → People → Pets → Things — verify brackets show/hide per category
- [ ] With "People" filter active, tap on a non-person bracket → verify it doesn't lock
- [ ] Volume-down with "People" filter → verify it only locks on people
- [ ] Verify zoom pill shows during tracking and after pinch gesture
- [ ] Verify stealth mode still works (volume-up)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -60,6 +60,7 @@ dependencies {
     implementation(libs.androidx.ui.graphics)
     implementation(libs.androidx.ui.tooling.preview)
     implementation(libs.androidx.material3)
+    implementation(libs.androidx.material.icons)
     debugImplementation(libs.androidx.ui.tooling)
 
     // CameraX
--- a/app/src/main/java/com/haptictrack/tracking/TrackingState.kt
+++ b/app/src/main/java/com/haptictrack/tracking/TrackingState.kt
@@ -15,6 +15,13 @@ enum class CaptureMode {
     PHOTO
 }
 
+enum class TrackingFilter {
+    ALL,
+    PERSON_ONLY,
+    PETS,
+    NON_PERSON_ONLY
+}
+
 data class TrackedObject(
     val id: Int,
     val boundingBox: RectF,
@@ -80,5 +87,7 @@ data class TrackingUiState(
     /** OIS compensation active (scale correction to avoid overcorrecting). */
     val oisCompensation: Boolean = true,
     /** Optical-flow translation correction on top of gyro rotation EIS. */
-    val translationEis: Boolean = true
+    val translationEis: Boolean = true,
+    /** Which object categories to show and allow tracking. */
+    val trackingFilter: TrackingFilter = TrackingFilter.ALL
 )
--- a/app/src/main/java/com/haptictrack/tracking/TrackingState.kt
+++ b/app/src/main/java/com/haptictrack/tracking/TrackingState.kt
@@ -22,6 +22,18 @@ enum class TrackingFilter {
     NON_PERSON_ONLY
 }
 
+private val ANIMAL_LABELS = setOf(
+    "cat", "dog", "bird", "horse", "sheep", "cow",
+    "elephant", "bear", "zebra", "giraffe"
+)
+
+fun labelMatchesFilter(label: String?, filter: TrackingFilter): Boolean = when (filter) {
+    TrackingFilter.ALL -> true
+    TrackingFilter.PERSON_ONLY -> label == "person"
+    TrackingFilter.PETS -> label in ANIMAL_LABELS
+    TrackingFilter.NON_PERSON_ONLY -> label != "person"
+}
+
 data class TrackedObject(
     val id: Int,
     val boundingBox: RectF,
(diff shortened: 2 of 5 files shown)


tuky191 (commented):
## PR Review: Camera UI redesign + tracking filter

Good work -- clean rewrite that transforms the debug UI into a professional camera interface. The component decomposition is well done, animations are tasteful, and the filter logic is consistent across all entry points.

---

### [minor] Duplicated filter logic -- `matchesFilter` vs `labelMatchesFilter`

`CameraViewModel.kt` (line 418) defines `matchesFilter()` with `ANIMAL_LABELS`, and `CameraScreen.kt` (line 796) defines `labelMatchesFilter()` with `ANIMAL_LABELS_UI`. They are identical implementations with different names and separate label sets. If the animal list ever changes, one will get out of sync.

Consider extracting a single `matchesFilter` function (or making it a method on `TrackingFilter`) in `TrackingState.kt` alongside the enum, so both files share one source of truth.

### [minor] `NON_PERSON_ONLY` ("Things") includes animals

`label != "person"` means the "Things" filter shows cats, dogs, etc. alongside chairs and cups. This may be intentional (keep it simple), but it's worth a conscious decision -- users selecting "Things" probably expect inanimate objects. If you want to exclude animals: `label != "person" && label !in ANIMAL_LABELS`.

### [minor] Dead intermediate variable in ViewModel

Lines 117-124 in `CameraViewModel.kt`:
```kotlin
val rawDisplayObjects = if (status == TrackingStatus.IDLE) { ... }
val displayObjects = rawDisplayObjects
```
The rename from `displayObjects` to `rawDisplayObjects` with a passthrough assignment looks like a leftover from a removed filtering step. Can collapse back to `val displayObjects = if (status == ...) { ... }`.

### [nit] `TopBar` zoom pill conditional creates layout shift

In `TopBar`, the zoom pill is conditionally included with `if (zoomAlpha > 0.01f)`. When it appears/disappears, the Row's center content changes and the recording indicator / settings button may shift slightly since the zoom pill doesn't occupy space when hidden. This is mitigated by the `weight(1f)` on left/right boxes, but the center content still pops in/out rather than fading. Consider always rendering the pill but with `Modifier.alpha(zoomAlpha)` when hidden, to avoid any layout jump.

### [nit] Settings gear always visible

The settings gear is always visible (even while recording or tracking). This is fine for dev/debug access, but on the final product you might want to hide it during recording to prevent accidental taps changing EIS settings mid-shot. No change needed now.

### [nit] `ModalBottomSheet` without explicit `sheetState`

The `ModalBottomSheet` uses the default `rememberModalBottomSheetState()` internally. This works fine, but if you ever need to control the sheet programmatically (e.g., dismiss on a specific event), you'd need to hoist the state. Fine for now.

### Correctness checks -- all pass

- **Filter consistency**: `onTapToLock` (line 182), `onVolumeDown` (line 260), and `TrackingOverlay` (line 824) all apply the filter. `clearTracking` preserves `trackingFilter` (line 369). The filter is only shown/changeable when idle + not recording (line 575). All correct.
- **State preservation**: `clearTracking()` correctly preserves `trackingFilter` in the new `TrackingUiState`.
- **No dead imports**: All old imports (`Path`, `PathEffect`, `StrokeCap`, `StrokeJoin`, `Button`, `ButtonDefaults`, `BoxWithConstraints`, `MaterialTheme`, `PaddingValues`, `LocalDensity`, `asAndroidPath`, `Stroke`) are properly removed. New imports are all used.
- **Recording timer**: Correctly resets on stop, uses `mutableIntStateOf` (the right primitive-optimized state for Int).
- **`scaledStroke`**: Still used by `drawIdleBrackets` -- not dead code.
- **`drawBracketLines`**: `pathEffect` parameter correctly removed (was only used by the now-deleted `drawDashedBrackets`).
- **No race conditions**: Filter reads from `_uiState.value` which is fine -- `cycleTrackingFilter` is UI-thread only.

### Verdict

No critical or major issues. The duplicated filter logic is the main thing worth fixing before it drifts -- everything else is nits. LGTM.

PR #173, opened 2026-05-07
Haptic heartbeat: pulse-rate centering + axis hint

tuky191:
## Summary
- Replace per-frame one-shot vibrations with a coroutine-driven heartbeat loop
- Pulse rate (900ms→180ms) and amplitude (20→180) encode centering quality
- Single pulse = horizontal drift, double pulse = vertical drift (axis hint)
- Lost/searching: distinct slow double-tap every 1.5s
- Dead zone 0.05, full urgency by 0.50 — tuned for zoomed subjects

## Test plan
- [ ] Lock on object centered → feel slow gentle pulses (~1Hz)
- [ ] Pan horizontally off-center → single pulses speed up
- [ ] Pan vertically off-center → double pulses speed up
- [ ] Near frame edge → rapid strong pulses
- [ ] Lose tracking → distinct slow double-tap pattern
- [ ] Clear tracking → silence
- [ ] With auto-zoom active, verify feedback reacts before target reaches edge

### DIFF
--- a/app/src/main/java/com/haptictrack/camera/CameraManager.kt
+++ b/app/src/main/java/com/haptictrack/camera/CameraManager.kt
@@ -212,7 +212,7 @@ class CameraManager(private val context: Context) {
                 onFrame = { bitmap -> onAnalysisFrame?.invoke(bitmap) },
                 onViewfinderFrame = { bitmap -> onViewfinderFrame?.invoke(bitmap) },
                 stabMatrixProvider = { gyroStabilizer.getMatrix() },
-                onRawFrame = { bitmap -> gyroStabilizer.onRawFrame(bitmap) }
+                onRawFrame = { bitmap, ts -> gyroStabilizer.onRawFrame(bitmap, ts) }
             )
             val readerSurface = reader.start()
             frameReader = reader
--- a/app/src/main/java/com/haptictrack/camera/GyroStabilizer.kt
+++ b/app/src/main/java/com/haptictrack/camera/GyroStabilizer.kt
@@ -55,6 +55,9 @@ class GyroStabilizer(context: Context) : SensorEventListener {
         // with matching -3dB cutoff: f_c = 1/(2πTC) for one pole, ≈0.187/σ for Gaussian.
         private const val FAST_SIGMA_TC_SCALE = 1.175
         private const val LOCAL_VEL_WINDOW_NS = 60_000_000L  // ±60ms for frame-local velocity
+        // Budget-aware TC: fraction of the leash budget the smoother may demand
+        // (deviation ≈ rate × TC must fit cropMargin/fEff at the current zoom).
+        private const val TC_BUDGET_SAFETY = 0.8
 
         /**
          * Per-device correction for intrinsics that don't match the HAL's actual
@@ -456,7 +459,7 @@ class GyroStabilizer(context: Context) : SensorEventListener {
                 it.println("frame_idx,timestamp_ns")
             }
             benchCorrWriter = PrintWriter(FileWriter(File(dir, "corrections.csv")), false).also {
-                it.println("frame_idx,timestamp_ns,raw_w,raw_x,raw_y,raw_z,smooth_w,smooth_x,smooth_y,smooth_z,eff_tc,corr_deg,leash,m0,m1,m2,m3,m4,m5,m6,m7,m8,zoom")
+                it.println("frame_idx,timestamp_ns,raw_w,raw_x,raw_y,raw_z,smooth_w,smooth_x,smooth_y,smooth_z,eff_tc,corr_deg,leash,m0,m1,m2,m3,m4,m5,m6,m7,m8,zoom,trans_cum_x,trans_cum_y,trans_vid_x,trans_vid_y")
             }
             PrintWriter(FileWriter(File(dir, "bench_params.csv"))).use { pw ->
                 pw.println("timeConstant,cropZoom,fxUv,fyUv,clampMarginFraction,oisCompensation")
@@ -487,11 +490,13 @@ class GyroStabilizer(context: Context) : SensorEventListener {
         val sq = smoothedQuat
         val mat = currentMatrix.get()
         val leash = if (lastLeashActive) 1 else 0
+        val videoTrans = videoTranslationCorrection(timestampNs)
         cw.println("$frameIdx,$timestampNs," +
             "${rq.w},${rq.x},${rq.y},${rq.z}," +
             "${sq.w},${sq.x},${sq.y},${sq.z}," +
             "$effectiveTc,$lastCorrAngleDeg,$leash," +
-            "${mat[0]},${mat[1]},${mat[2]},${mat[3]},${mat[4]},${mat[5]},${mat[6]},${mat[7]},${mat[8]},$zoomRatio")
+            "${mat[0]},${mat[1]},${mat[2]},${mat[3]},${mat[4]},${mat[5]},${mat[6]},${mat[7]},${mat[8]},$zoomRatio," +
+            "$transCumX,$transCumY,${videoTrans.first},${videoTrans.second}")
     }
 
     /** Get the current stabilization matrix (column-major mat3, 9 floats). Thread-safe. */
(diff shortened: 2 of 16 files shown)


tuky191 (commented):
## Review: Haptic heartbeat — pulse-rate centering + axis hint

Good concept — replacing per-frame one-shot vibrations with a steady heartbeat loop is the right direction for hands-free usability. The axis hint (single vs double pulse) is a nice touch. A few issues to address before merging:

---

### [critical] Unscoped CoroutineScope leaks if shutdown() is never called

`HapticFeedbackManager` creates a bare `CoroutineScope(Dispatchers.Default)` that is never structurally cancelled. `shutdown()` cancels the current `heartbeatJob`, but the scope itself lives forever. If `onCleared()` runs but a new `heartbeatJob` is launched between `hapticManager.shutdown()` and GC (unlikely but possible in a race), that coroutine has no parent to cancel it.

More practically: if `shutdown()` is skipped for any reason (crash, early return), the scope leaks its thread. The fix is straightforward — use a `SupervisorJob`-backed scope and cancel the scope itself in `shutdown()`:

```kotlin
private val scope = CoroutineScope(Dispatchers.Default + SupervisorJob())

fun shutdown() {
    scope.cancel()   // cancels heartbeatJob + prevents future launches
    vibrator.cancel()
}
```

Alternatively, accept the ViewModel's `viewModelScope` as a constructor parameter so the lifecycle is managed externally.

---

### [major] updateTrackingStatus is not thread-safe — torn reads on status transition

`updateTrackingStatus` is called from the detection callback (background thread). It writes `driftX`, `driftY`, then checks `status == currentStatus`, then writes `currentStatus`, then cancels and relaunches the job. Multiple fields are written and read without synchronization beyond `@Volatile` on individual fields.

The race: two rapid calls (e.g., LOCKED then LOST) can interleave such that the second call's `heartbeatJob?.cancel()` runs before the first call's `scope.launch` assigns to `heartbeatJob`, leaving two coroutines running simultaneously.

Concrete scenario:
1. Thread A enters `updateTrackingStatus(LOCKED, ...)`, passes the `status == currentStatus` check, sets `currentStatus = LOCKED`, calls `heartbeatJob?.cancel()` (cancels nothing — null).
2. Thread B enters `updateTrackingStatus(LOST, ...)`, passes the check (LOST != LOCKED), sets `currentStatus = LOST`, calls `heartbeatJob?.cancel()` (still null — Thread A hasn't assigned yet).
3. Thread A assigns `heartbeatJob = scope.launch { lockedHeartbeat() }`.
4. Thread B assigns `heartbeatJob = scope.launch { lostHeartbeat() }`.
5. Now both `lockedHeartbeat()` AND `lostHeartbeat()` are running. The old locked job reference is lost and cannot be cancelled.

Fix: synchronize the entire method body, or confine state mutation to a single thread (e.g., `Dispatchers.Main.immediate`). Since this is haptics, `Main` dispatch latency is fine:

```kotlin
private val scope = CoroutineScope(Dispatchers.Main.immediate + SupervisorJob())

fun updateTrackingStatus(status: TrackingStatus, driftX: Float = 0f, driftY: Float = 0f) {
    // Now single-threaded — no races
    this.driftX = driftX
    this.driftY = driftY
    if (status == currentStatus) return
    // ...
}
```

Or if you want to keep `Dispatchers.Default` for the heartbeat, use a `Mutex` or `@Synchronized` on the method.

---

### [major] driftX/driftY updates are silently dropped when status hasn't changed

When `status == currentStatus` (the common case — LOCKED stays LOCKED every frame), the method returns immediately after writing `driftX`/`driftY`. The coroutine picks up the new values on its next loop iteration, which is correct. **But** if the detection callback fires faster than the heartbeat interval (e.g., 30fps vs 900ms heartbeat at center), the coroutine reads stale drift values for up to 900ms.

This means haptic feedback can lag almost a full second behind the actual object position when centered (low urgency = long interval). If the user quickly pans off-center, the urgency ramp won't kick in until the current sleep expires.

This is major because 900ms of haptic lag at center feels unresponsive. Consider a shorter sleep granularity (e.g., 50ms ticks with a modulo counter for pulse timing), so drift changes are reflected within one tick.

---

### [minor] Dead code: `ZoomController.calculateEdgeProximity` is now unused

The old `edgeProximity` call site was replaced by inline drift computation. `calculateEdgeProximity` in `ZoomController.kt` (line 115) is now dead code. Clean it up in this PR or file a follow-up.

---

### [minor] Double-pulse total cycle time is ~40% longer than single-pulse at same urgency

In the double-pulse path:
```kotlin
vibrator.vibrate(VibrationEffect.createOneShot(pulseDuration, tapAmp))
delay(pulseDuration + 60)
vibrator.vibrate(VibrationEffect.createOneShot(pulseDuration, amplitude))
delay(intervalMs)
```

Total cycle = `2*pulseDuration + 60 + intervalMs`. At max urgency: `2*70 + 60 + 180 = 380ms` vs single-pulse `70 + 180 = 250ms`. The double-pulse path runs ~40% slower, making vertical drift feel less urgent than horizontal drift at the same edge proximity. If intentional, add a comment; otherwise subtract the extra time from `intervalMs`.

---

### [nit] `while (true)` — prefer `while (isActive)` for explicit cancellation contract

`delay()` is cancellable so this works correctly, but `while (isActive)` makes the contract explicit and prevents accidentally adding non-cancellable code later.

---

### [nit] Magic numbers for urgency dead zone and saturation should be named constants

`0.05f` (dead zone) and `0.45f` (saturation range) are tunable parameters. Per project convention (see the Tunable Parameters table in CLAUDE.md), these should be named companion object constants.

---

**Summary:** The architecture is sound and the UX concept is good. The thread-safety race (two heartbeats running simultaneously) and the leaked CoroutineScope are the blockers — both have straightforward fixes. The drift-lag concern is worth thinking about but could ship as-is with a documented tradeoff.

tuky191:
## PR #173 Review: Geiger counter haptic feedback

### Thread safety

The concurrency model is sound. `@Synchronized` on `updateTrackingStatus` serializes status transitions and job cancel/launch. `@Volatile` on `driftX`/`driftY`/`strength` ensures the `geigerLoop` coroutine (running on `Dispatchers.Default`) sees fresh values each iteration without needing a lock. `heartbeatJob` is only accessed inside the synchronized block, so no annotation needed there.

`SupervisorJob()` + `scope.cancel()` in `shutdown()` is correct for lifecycle.

### LOST/SEARCHING/IDLE silence

Correct. Lines 55-63: on any transition away from LOCKED, `heartbeatJob?.cancel()` + `vibrator.cancel()` fire, and no new job is launched. The early return on line 52 means repeated IDLE/LOST calls are no-ops after the first transition. No issues here.

### Issues

**[major] `HapticFeedbackManager.kt` line 78: strength=0 still vibrates (slider says "Off" but minimum amplitude is 1)**

When the user drags the haptic strength slider to "Off" (0.0), the code computes `amp = (0.0f * 255f).toInt().coerceIn(1, 255)` which equals 1. The `coerceIn(1, 255)` floor prevents true silence. The slider label "Off" on `CameraScreen.kt` line 1014 promises silence at the left end, but the user will feel faint clicks.

Fix: skip the vibrate call entirely when `strength <= 0`:
```kotlin
val amp = (strength * 255f).toInt().coerceIn(1, 255)
if (strength <= 0f) {
    delay(intervalMs)
    continue
}
```
Or alternatively, guard both the single-click and double-click branches with the same check.

**[minor] `HapticFeedbackManager.kt` lines 80-89: double-click cycle is ~25ms longer than single-click at the same urgency**

Single-click cycle (start-to-start): the vibrate is async, so the cycle is approximately `intervalMs`.

Double-click cycle: `(PULSE_MS + DOUBLE_PULSE_GAP_MS)` wait + second vibrate + `max(intervalMs - PULSE_MS - DOUBLE_PULSE_GAP_MS, 30)` wait = `intervalMs + PULSE_MS` when `intervalMs >= 70`. That's 25ms extra per cycle.

At the fastest rate (120ms), single clicks fire every ~120ms but double clicks every ~145ms -- a 21% difference. The second commit message says this was fixed ("equal pulse timing"), but the third commit's rewrite reintroduced it.

Fix: subtract both pulse durations from the final delay:
```kotlin
delay(maxOf(intervalMs - 2 * PULSE_MS - DOUBLE_PULSE_GAP_MS, 30L))
```

**[minor] `HapticFeedbackManager.kt` line 80: centering < 0.95f threshold for double-click is arbitrary and undocumented**

When the object is nearly centered (centering >= 0.95, meaning edgeProximity < 0.0725), vertical drift never produces a double-click even if `dy > dx`. This dead zone for axis-hinting is separate from `DEAD_ZONE` and should probably be a named constant for discoverability:
```kotlin
private const val AXIS_HINT_THRESHOLD = 0.95f
```

**[nit] `CameraScreen.kt` HAPTICS section: no enable/disable toggle**

The gyro EIS section has a master toggle that hides sub-settings when off. The haptics section only has a strength slider. Consider adding an on/off `SettingRow` so the user can disable haptics entirely with a single tap rather than dragging the slider to zero (especially since zero doesn't actually silence it per the major issue above).

**[nit] `CameraViewModel.kt` line 90-95: drift computation repeated inline**

The `driftX`/`driftY` calculation `(centerX - 0.5f) * 2f` maps normalized [0,1] coords to [-1,1]. This is fine, but the `* 2f` means an object at the frame edge (center at 0.0 or 1.0) produces drift = -1.0 or +1.0. With `DEAD_ZONE=0.05` and `URGENCY_RANGE=0.45`, full urgency (centering=0) is reached at `edgeProximity=0.50`, i.e., when the object center is at 0.75 normalized. That means the outer 25% of the frame produces no additional urgency change -- it's already at max-slow clicks. This is probably intentional for zoomed subjects, but worth documenting in a comment.

### Summary

The architecture is clean -- coroutine-driven heartbeat, volatile drift reads, synchronized transitions. The critical fix is the strength=0 silence bug. The timing discrepancy is minor but audible at high urgency.

